PR #61, opened 2017-03-30
Update install instructions

jimthedev:
Add a bit more information on the expectations of how a haul works with a react-native project.


grabbou:
Thanks!

Issue #63, opened 2017-03-31
Doesn't work on Node 4.x


colinramsay:
I tried running `haul` with v4.6.1 of `node` and it failed (`ReferenceError: regeneratorRuntime is not defined`), but with v6 it works. The readme should indicate the minimum required version.

grabbou:
Can you try running #64 to see if it fixes your problem? We support Node 4, but our preset is not including regenerator-runtime. Should be fixed after this PR gets merged and released.

Until then, you can point `haul-cli` to `callstack-io/haul#1afd58644ad576376d481eb6dbf70ea291bf269a` and run install again.

colinramsay:
Oh I just used `nvm` to switch to v6, it wasn't a major issue for me. I just thought it would be useful for the readme to be explicit about it :)

grabbou:
@colinramsay we do support v4, so thanks to your report we were able to locate an issue with our babel preset!

colinramsay:
:man_dancing:

grabbou:
0.1.4 should now work on Node 4

PR #64, opened 2017-03-31
Include polyfills with babel-preset-env

grabbou:
Fixing #63 as per this comment https://github.com/babel/babel-preset-env/issues/92#issuecomment-266869180

Babel-preset-env will use `regenerator-runtime` on Node 4, but we don't include polyfill from it. Turning on settings as in this PR will conditionally turn on polyfills as needed.

### DIFF
--- a/bin/cli
+++ b/bin/cli
@@ -1,3 +1,9 @@
 #!/usr/bin/env node
 
+require('babel-register')({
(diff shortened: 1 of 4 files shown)


grabbou:
Needs further testing on Node v4.6.0 - there's another error in `src/utils/polyfillEnvironment.js`

PR #68, opened 2017-03-31
Correct configuration docs

martypenner:
This one had me puzzled. I followed the configuration docs to start customizing my project. I copy/pasted the last code example, changed the `rules` key to be an array (as webpack requires), and removed the custom bits in order to test:

```
  entry: `./index.${platform}.js`,
  module: {
    ...defaults.module,
    rules: [
      ...defaults.rules
    ]
  },
  resolve: {
    ...defaults.resolve
  }
});
```

I then received the following error:

screen shot 2017-03-31 at 6 18 30 am

Turns out the `...defaults.rules` spread should actually be `...defaults.module.rules`. Took me a while to figure that out :)

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -41,13 +41,13 @@ module.exports = ({ platform }, defaults) => ({
   entry: `./index.${platform}.js`,
   module: {
     ...defaults.module,
-    rules: {
+    rules: [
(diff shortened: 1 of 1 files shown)


grabbou:
Great! Thanks for fixing the readme for others :)

PR #72, opened 2017-03-31
Init command to add entries related to Haul in .gitignore.

ferrannp:


src/commands/init.js
```diff
@@ -96,6 +96,25 @@ async function init() {
     }
   }
 
+  progress = ora('Adding new entries to .gitignore').start();
+
+  const pathToGitIgnore = path.join(cwd, '.gitignore');
+  if (fs.existsSync(pathToGitIgnore)) {
+    const _gitignore = fs.readFileSync(pathToGitIgnore);
```
satya164: nit: remove underscore

src/commands/init.js
```diff
@@ -96,6 +96,25 @@ async function init() {
...
+      const buffer = new Buffer('# Haul\n#\n.happypack\nhaul-debug.log\n\n');
+      fs.writeSync(fd, buffer, 0, buffer.length, 0);
+      fs.writeSync(fd, _gitignore, 0, _gitignore.length, buffer.length);
+      fs.close(fd);
+      progress.succeed('Added new entries to .gitignore');
+    } else {
+      progress.info('.gitignore already contains # Haul');
+    }
+  } else {
+    progress.warn('No .gitignore found, skipping adding new entries');
```
satya164: `No .gitignore found, skipping`

src/commands/init.js
```diff
@@ -96,6 +96,25 @@ async function init() {
     }
   }
 
+  progress = ora('Adding new entries to .gitignore').start();
+
+  const pathToGitIgnore = path.join(cwd, '.gitignore');
+  if (fs.existsSync(pathToGitIgnore)) {
+    const _gitignore = fs.readFileSync(pathToGitIgnore);
+    if (!_gitignore.includes('# Haul')) {
```
satya164: prolly should check for individual entries rather than the comment?

src/commands/init.js
```diff
@@ -96,6 +96,25 @@ async function init() {
...
+      const fd = fs.openSync(pathToGitIgnore, 'w+');
+      const buffer = new Buffer('# Haul\n#\n.happypack\nhaul-debug.log\n\n');
+      fs.writeSync(fd, buffer, 0, buffer.length, 0);
```
satya164: nit: maybe a new line here for better readability

src/commands/init.js
satya164: nit: new line before

src/commands/init.js
grabbou: Maybe we create it if it's missing?

src/commands/init.js
```diff
@@ -96,6 +96,25 @@ async function init() {
...
+      fs.writeSync(fd, _gitignore, 0, _gitignore.length, buffer.length);
+      fs.close(fd);
+      progress.succeed('Added new entries to .gitignore');
+    } else {
+      progress.info('.gitignore already contains # Haul');
```
grabbou: I'd rephrase that to something better, e.g. `Haul files are already ignored`

src/commands/init.js
```diff
@@ -96,6 +96,25 @@ async function init() {
     }
   }
 
+  progress = ora('Adding new entries to .gitignore').start();
```
grabbou: How about we show this only if: a) `gitignore` exists and it doesn't contain `haul` yet? Otherwise it feels like unnecessary output.

src/commands/init.js
satya164: In that case, we should detect if it's a git project first

src/commands/init.js
grabbou: +1

src/commands/init.js
ferrannp: In theory, it should be there if its a RN app. If it is not, that is probably dev decision so I would not create it or anything.

satya164: Sounds fair

src/commands/init.js
```diff
@@ -96,6 +96,25 @@ async function init() {
...
+
+  const pathToGitIgnore = path.join(cwd, '.gitignore');
+  if (fs.existsSync(pathToGitIgnore)) {
+    const _gitignore = fs.readFileSync(pathToGitIgnore);
+    if (!_gitignore.includes('# Haul')) {
+      const fd = fs.openSync(pathToGitIgnore, 'w+');
+      const buffer = new Buffer('# Haul\n#\n.happypack\nhaul-debug.log\n\n');
```
satya164: Why buffer instead of just appending the text to what we read above?

ferrannp: To add it at the beginning of the file, not at the end.

satya164: Then you can just prepend, no? Though I think appending is better

```
fs.writeFileSync(dedent`${gitignore}

# Haul
haul-debug.log
.happypack
`, pathToFile);

File: src/commands/init.js [hunk 8]
```diff
@@ -96,23 +96,33 @@ async function init() {
...
-    if (!_gitignore.includes('# Haul')) {
-      const fd = fs.openSync(pathToGitIgnore, 'w+');
-      const buffer = new Buffer('# Haul\n#\n.happypack\nhaul-debug.log\n\n');
-      fs.writeSync(fd, buffer, 0, buffer.length, 0);
-      fs.writeSync(fd, _gitignore, 0, _gitignore.length, buffer.length);
-      fs.close(fd);
-      progress.succeed('Added new entries to .gitignore');
+    const gitignore = fs.readFileSync(pathToGitIgnore);
+
+    if (!gitignore.includes('# Haul')) {
```
satya164: `gitignore.split('\n').includes`

grabbou:
What's the status of this PR?

ferrannp:
I need to finish with the split stuff, I'll let you know.

PR #74, opened 2017-03-31
Adds symbolication support.

skellock:
React Native jumps back to the "packager" for source maps.

This allows us to have frames that look like this:

image

Instead of `index.ios.bundle?line=69696`.

I needed this for Reactotron, so we can print some prettier messages.

image

Just a heads up, you'll see a really horrible hack I had to do to get this to work in `polyfillEnvironment.js`.

Thank you SO much for making this project.  I'm using symlinks again on my react native 3rd party library!  😻


satya164:
This is awesome! Thanks. Gonna have a look in the morning.

grabbou:
This is great! Thanks for trying out and adding support for this! Will also review in the morning 👌

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -0,0 +1,128 @@
...
+const path = require('path');
+const delve = require('dlv');
+const messages = require('../../messages');
+const logger = require('../../logger');
+
+/**
+ * Create an Express middleware for handling React Native symbolication requests
+ *
+ * @param {*} compiler - WebPack compiler.
+ * @return {function} The Express middleware.
```
grabbou: Since we use `Flow` (and we could try to do it in this file as well), I believe we can remove type annotations like that from files

grabbou: We use `Flow` in other files, so I am thinking instead of type annotations we could use Flow instead?

skellock:
@grabbou All flowed up now.  

I'm new to Flow, so if I've done something pants-on-head wrong, please lemme know.

A few notes:

* I decided to keep the types local to the file as opposed to piling onto `types.js`. I'm not sure they'll be used anywhere else, but if you want me relocate them, I'm cool with that.
* I `flow-typed` to bring in `express`, but when I did, there were a few problems popped up in the codebase (`new Express()` is't a thing it told me).  If you're interested, I could do this in a new PR.
* There isn't a `flow-typed` entry for `source-map`. It said I should create a stub, but I pressed control-d instead.  👀

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -0,0 +1,163 @@
...
+const path = require('path');
+const delve = require('dlv');
+const messages = require('../../messages');
+const logger = require('../../logger');
+
+/**
+ * React Native's version of a stack frame.
+ */
+type ReactNativeStackFrame = {
+  // ths
```
satya164: nit: what does this comment mean? maybe unnecessary?

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -0,0 +1,163 @@
...
+ */
+
+const SourceMapConsumer = require('source-map').SourceMapConsumer;
+const path = require('path');
+const delve = require('dlv');
+const messages = require('../../messages');
+const logger = require('../../logger');
+
+/**
+ * React Native's version of a stack frame.
```
satya164: nit: comment seems unnecessary since variable says the same thing

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -0,0 +1,163 @@
...
+type ReactNativeStackFrame = {
+  // ths
+  lineNumber: number,
+  column: number,
+  file: string,
+  methodName: string,
+};
+
+/**
+ * A list of React Native stack frames.
```
satya164: nit: Unnecessary comment

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -0,0 +1,163 @@
...
+
+  /**
+   * The Express middleware for symbolicatin'.
+   */
+  function symbolicateMiddleware(
+    req: any,
+    res: any,
+    next: Function,
+  ): ?Function {
+    // jet unless we're asked to symbolicate
```
satya164: nit: what does `jet` mean here?

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -0,0 +1,163 @@
...
+};
+
+/**
+ * Create an Express middleware for handling React Native symbolication requests
+ */
+function create(compiler: any): Function {
+  /**
+   * Creates a SourceMapConsumer so we can query it.
+   */
+  function createSourceMapConsumer() {
```
satya164: nit: let's move this function outside to avoid extra indentation

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -0,0 +1,163 @@
...
+    // In haul, webpack stores its bundled stuff in a memory file system. Let's
+    // go grab the root object where we expect the source map to live.
+    const hops: Array<string> = compiler.options.output.path
+      .split(path.sep)
+      .filter(x => x !== '');
```
satya164: question: what's `x` here and what does it mean when `x !== ''`?

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -0,0 +1,163 @@
...
+ * The payload that is returned to React Native after we symbolicate.
+ */
+type ReactNativeSymbolicateResponse = {
+  stack: ReactNativeStack,
+};
+
+/**
+ * Create an Express middleware for handling React Native symbolication requests
+ */
+function create(compiler: any): Function {
```
satya164: nit: do we need to annotate as `any`? maybe `*` is better?

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -0,0 +1,163 @@
...
+    const hops: Array<string> = compiler.options.output.path
+      .split(path.sep)
+      .filter(x => x !== '');
+    const base = delve(compiler.outputFileSystem.data, hops);
+
+    // grab the Buffer for the source map
+    const sourceMapBuffer = base &&
+      base[`${compiler.options.output.filename}.map`];
+
+    // jet if it's missing?
```
satya164: does `jet` mean `return`?

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -0,0 +1,163 @@
...
+    } catch (err) {
+      logger.error(messages.sourceMapInvalidFormat());
+      return null;
+    }
+  }
+
+  /**
+   * Gets the stack frames that React Native wants us to convert.
+   */
+  function getRequestedFrames(req: any): ?ReactNativeStack {
```
satya164: I'd move this function outside to avoid extra indentation

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -0,0 +1,163 @@
...
+      // should happen, but at least we won't die
+      return null;
+    }
+  }
+
+  /**
+   * The Express middleware for symbolicatin'.
+   */
+  function symbolicateMiddleware(
+    req: any,
```
satya164: maybe `*` instead of `any` is better

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -0,0 +1,163 @@
...
+    } catch (err) {
+      // should happen, but at least we won't die
+      return null;
+    }
+  }
+
+  /**
+   * The Express middleware for symbolicatin'.
+   */
+  function symbolicateMiddleware(
```
satya164: I'd move this function outside to avoid extra indentation

skellock:
Cheers thx for the review @satya164 .  Made those changes.  Jet means to leave ya... sorry about that.  Slang.  haha.

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Create an Express middleware for handling React Native symbolication requests
+ */
+function create(compiler: *): Function {
+  /**
+   * The Express middleware for symbolicatin'.
+   */
+  function symbolicateMiddleware(req: *, res: *, next: Function): ?Function {
+    if (req.url !== '/symbolicate') return next();
```
satya164: Probably we should check for `req.path` instead?

satya164:
Thank you! Sorry, I missed one.

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -0,0 +1,147 @@
...
+        };
+      },
+    );
+
+    // send it back to React Native
+    const responseObject: ReactNativeSymbolicateResponse = {
+      stack: convertedFrames,
+    };
+    const response = JSON.stringify(responseObject);
+    res.end(response);
```
grabbou: We should probably use `res.json(responseObject)` instead?

PR #76, opened 2017-03-31
Tiny Grammar Problem

CL123123:
In the first paragraph, there is a tiny grammar problem.


satya164:
@grabbou I cannot merge PRs due to cricle ci

Issue #87, opened 2017-04-02
Add script to build process during init


satya164:
Should replace React Native's packager CLI in gradle and xcode scripts with haul during init.

- [x] iOS ~ @grabbou
- [ ] Android ~ @satya164

grabbou:
There's a problem on first Xcode run as `haul-integrate.sh` will be executed after React own build scripts are executed. That means the very first run will launch packager no matter what we do. Second works properly. Figure out how to fix it. 

We might need to run `haul-integrate.sh` during `init` command to make the first change. The purpose of it to sit in `build script` is to avoid updates of React / npm installs.

grabbou:
Now that we have ENV variables, we will be removing this legacy workaround. See #234 for details.

PR #90, opened 2017-04-03
Add iOS build phase

grabbou:
Issue #87 

As of now, the only valid approach for integrating Haul with existing React Native applications is to define a script phase that runs every time we build our project and overwrites `packager.sh` (used to automatically start packager on run) and `react-native-xcode.sh` (used to bundle your app).

Problem with setting env variable `CLI_PATH` is that some scripts are defined in `React.xcodeproj`. That means scripts defined there will not receive env variables set by parent project.

Thanks to regular expressions, this is rather easy.


grabbou:
screen shot 2017-04-03 at 10 17 04

grabbou:
I decided to vendor `react-native-xcode.sh` until it ships with React Native. When that happens, we should detect version and either use vendor copy or directly from `react-native`. This is to make it possible for older apps to `bundle`.

src/commands/init.js
```diff
@@ -140,6 +142,122 @@ async function init() {
...
+    }
+  }
+
+  // Otherwise, ask for path to a file
+  if (!entry) {
+    const result = await inquirer.prompt([
+      {
+        type: 'input',
+        name: 'entry',
+        message: 'Enter the name of the .xcodeproj file',
```
satya164: Isn't this the `path` and not `name`?

grabbou: Yup, it's path

src/commands/init.js
```diff
@@ -140,6 +142,122 @@ async function init() {
...
+  }
+
+  const progress = ora('Adding haul to your native build pipeline');
+
+  await sleep();
+
+  let project = fs.readFileSync(path.join(entry, 'project.pbxproj')).toString();
+
+  // Are we already integrated?
+  if (project.includes('AD0CE2C91E925489006FC317')) {
```
satya164: Let's store this in a constant

src/commands/init.js
```diff
@@ -140,6 +142,122 @@ async function init() {
...
+          fs.existsSync(pathToFile) && pathToFile.includes('.xcodeproj')
+            ? true
+            : `${pathToFile} is not a valid .xcodeproj`,
+      },
+    ]);
+
+    entry = path.resolve(cwd, result.entry);
+  }
+
+  const progress = ora('Adding haul to your native build pipeline');
```
satya164: `Adding haul to your XCode build scripts`

Though I feel like we should ask before this change. Looks rather invasive and I might just want to try before integrating.

src/commands/init.js
```diff
@@ -140,6 +142,122 @@ async function init() {
...
+
+  await sleep();
+
+  let project = fs.readFileSync(path.join(entry, 'project.pbxproj')).toString();
+
+  // Are we already integrated?
+  if (project.includes('AD0CE2C91E925489006FC317')) {
+    progress.info('Haul is already part of your build pipeline');
```
satya164: nit: Mention XCode

src/commands/init.js
```diff
@@ -140,6 +142,122 @@ async function init() {
...
+        `,
+    );
+  });
+
+  if (sectionsCount > 0) {
+    fs.writeFileSync(path.join(entry, 'project.pbxproj'), project);
+    progress.succeed('Added haul to your native build pipeline');
+  } else {
+    progress.fail(
+      'Failed to add Haul to your build pipeline. See docs for manual instructions.',
```
satya164: nit: Mention XCode

Link to docs?

src/commands/init.js
```diff
@@ -140,6 +142,122 @@ async function init() {
...
+      dedent`
+        buildPhases = (
+          AD0CE2C91E925489006FC317 /* Integrate Haul with React Native */,
+        `,
+    );
+  });
+
+  if (sectionsCount > 0) {
+    fs.writeFileSync(path.join(entry, 'project.pbxproj'), project);
+    progress.succeed('Added haul to your native build pipeline');
```
satya164: nit: Mention XCode

src/commands/init.js
satya164: `build pipeline` sounds very serious, maybe `build scripts` is better

PR #91, opened 2017-04-03
Improvements to webpack config

satya164:


grabbou:
Looks good, how about we also add the plugin that allows you to `npm install` w/o restarting the server?

satya164:
Why would you need to restart server after npm install?

grabbou:
https://github.com/facebookincubator/create-react-app/blob/47a8148fb195707b4fb533521afd3aee6807d92a/packages/react-scripts/config/webpack.config.dev.js#L243

satya164:
Hmm looks like it's just that the dev server doesn't rebuild when a module changes inside node_modules and you have to save some file to trigger it, don't have to restart the server.

Though i don't like having a dependency on `create-react-app` to include the plugin it uses, probably we can vendor this one later.

grabbou:
Yeah, I wouldn't like to see `create-react-app` or `react-scripts` being pulled down just for the sake of that plugin.

We can probably talk with the maintainers later about pushing that plugin into separate `npm` module.

PR #97, opened 2017-04-04
Make loader platform specific

grabbou:
Fixes an issue when running with `--platform all` makes assets overwrite each other (Android overwrites iOS and then, iOS keeps returning Android back icon for instance).

### DIFF
--- a/src/loaders/assetLoader.js
+++ b/src/loaders/assetLoader.js
@@ -25,10 +25,14 @@ module.exports = async function assetLoader() {
   const dirname = path.dirname(filepath);
   const suffix = `(@\\d+(\\.\\d+)?x)?(\\.(${query.platform}|native))?\\.${info.type}$`;
   const filename = path.basename(filepath).replace(new RegExp(suffix), '');
-  const url = path.relative(query.root, dirname);
-  const longname = `${`${url.replace(/\//g, '_')}_${filename}`
-    .toLowerCase()
-    .replace(/[^a-z0-9_]/g, '')}.${info.type}`;
+
+  // When building for `all` platforms in dev mode (query.bundle == false)
+  // we need to make assets platform-specific, otherwise they overwrite
+  // each other
+  const url = path.relative(
+    query.root,
+    path.join(dirname, query.bundle ? '' : query.platform),
+  );
 
   const regex = new RegExp(`^${escapeStringRegexp(filename)}${suffix}`);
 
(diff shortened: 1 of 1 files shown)


grabbou:
This is not yet ready for merging as I am working on unit tests for that file & adding few more comments here and there.

Feedback appreciated.

satya164:
I think the simplest fix is to change `path.join('assets', url, name);` to `path.join('assets', query.bundle ? '' : query.platform, url, name);` and `path.join('/assets', url)` to `path.join('assets', query.bundle ? '' : query.platform, url))` (store the later in a variable and reuse)

This change changes the meaning of the `url` variable which was supposed to be the relative URL to the destination. It will likely break bundling assets for iOS. Not sure what else it will break since it was the original assumption while writing the code. 

Can we do the change I mentioned above instead? I'll add some comments to the code so it's easier to understand.

grabbou:
> This change changes the meaning of the url variable which was supposed to be the relative URL to the destination. It will likely break bundling assets for iOS. Not sure what else it will break since it was the original assumption while writing the code.

Not really sure, `url` is used only in two places, in those path.join calls. How concept of `relative url` fits in there?

satya164:
So it won't break anything else, but it'll definitely break bundling assets for iOS. Can we do the change I mentioned instead? I'll add comments later.

grabbou:
> So it won't break anything else, but it'll definitely break bundling assets for iOS

It shouldn't break bundling for any platform since we don't change the code when `query.bundle` is true. Your code suggestion is just moving where `path.join` happens.

src/loaders/assetLoader.js
```diff
@@ -111,18 +115,20 @@ module.exports = async function assetLoader() {
                   throw new Error(`Unknown scale ${scale} for ${filepath}`);
               }
 
+              const longname = `${`${url.replace(/\//g, '_')}_${filename}`
```
satya164: Why do this multiple times instead of once?

grabbou: Just wanted to scope it for now. Ultimately I want it to go to a separate function, so that will be gone.

satya164:
This will break iOS because iOS expects the assets to be folder inside `assets` which this PR removes.

src/loaders/assetLoader.js
satya164: Why does it need scoping? It's the same name for every file.

grabbou:
> This will break iOS because iOS expects the assets to be folder inside assets which this PR removes.

While bundling, Xcode specifies `--assetsDest` that already points to that folder.

src/loaders/assetLoader.js
grabbou: It doesn't. It's WIP and that was moved down to help me understand usage. We can ignore this line.

grabbou:
> This will break iOS because iOS expects the assets to be inside a folder assets which this PR removes.

@satya164 Yeah, I'll revert that change. Sorry for that. I had a folder `assets` already and it got me confused.

satya164:
Also I think `assets/ios/path/to/my/images/image.png` will be better rather than `path/to/my/images/ios/image.png`, because the former preserves the actual path to the file and prepends a path, as opposed to modifying the path.

satya164:
`url = path.join('assets', query.bundle ? '' : query.platform, path.relative(...))` seems much better

satya164:
Oh, `url` is also used in `longname`, so `url` needs to be the relative path.

grabbou:
I am going to resubmit it again in a clean state.

PR #98, opened 2017-04-04
Fix loader issues when running all platforms

grabbou:
Fixes an issue when running with --platform all makes assets overwrite each other (Android overwrites iOS and then, iOS keeps returning Android back icon for instance).

### DIFF
--- a/src/loaders/assetLoader.js
+++ b/src/loaders/assetLoader.js
@@ -23,9 +23,10 @@ module.exports = async function assetLoader() {
   const filepath = this.resourcePath;
   const info = size(filepath);
   const dirname = path.dirname(filepath);
-  const suffix = `(@\\d+(\\.\\d+)?x)?(\\.(${query.platform}|native))?\\.${info.type}$`;
+  const url = path.relative(config.root, dirname);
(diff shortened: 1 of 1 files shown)


src/loaders/assetLoader.js
```diff
@@ -132,7 +141,7 @@ module.exports = async function assetLoader() {
     }),
   );
 
-  let publicPath = `__webpack_public_path__ + ${JSON.stringify(path.join('/assets', url))}`;
+  let publicPath = `__webpack_public_path__ + ${JSON.stringify(path.join(assets, url))}`;
```
satya164: We will probably need to add a `/` before `JSON.stringify(path.join(assets, url))`

src/loaders/assetLoader.js
```diff
@@ -132,7 +141,7 @@ module.exports = async function assetLoader() {
     }),
   );
 
-  let publicPath = `__webpack_public_path__ + ${JSON.stringify(path.join('/assets', url))}`;
+  let publicPath = `__webpack_public_path__ + /${JSON.stringify(path.join(assets, url))}`;
```
satya164: this will be syntax error. we want  `path.join('/', assets, url)` instead

PR #99, opened 2017-04-05
Add to package.json scripts

### LINKED ISSUE #85: Add package.json scripts during init
Should check if no `start` script exists or it is the default React Native one, then prompt to add/replace it during init.

Otherwise prompt and add a script called `haul`

grabbou:
Fixes #85

The process is:

1) If Haul is already in `package.json` scripts (any of scripts contains haul), we print info that it's already there and how to run it

2) If there's start script defined already and is not a default packager, we ask for a name for Haul script. By default it's start so users can opt-in to overwrite it

3) We use either start or name provided in point 2 to write package.json.

Questions:

1) Currently we add `haul start`. As of now, one has to pass `-- --platform [ios|android|all]` in order to start Haul for these platforms. Shall we print that as well?

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -162,8 +162,9 @@ const addToPackageScripts = async (cwd: string) => {
 
   const scripts = pjson.scripts || {};
 
-  const haulScript = Object.keys(scripts).find(name =>
-    scripts[name].includes('haul'));
+  const haulScript = Object.keys(scripts).find(
+    name => scripts[name] === 'haul start',
+  );
 
   if (haulScript) {
     ora().info(
@@ -182,8 +183,8 @@ const addToPackageScripts = async (cwd: string) => {
       {
         type: 'input',
         name: 'scriptName',
-        message: 'Enter the name of the script to add to package.json',
-        default: scriptName,
+        message: 'Enter the name of the script to add to package.json, e.g. `haul` for `yarn run haul`',
+        default: 'haul',
       },
     ]);
 
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -167,7 +167,7 @@ const addToPackageScripts = async (cwd: string) => {
 
   if (haulScript) {
     ora().info(
-      `Haul is already in your package's scripts. You can start it by running '${getRunScript(haulScript)}'`,
+      `Haul already exists in your package.json. Start Haul by running ${getRunScript(haulScript)}'`,
     );
     return;
   }
@@ -182,7 +182,7 @@ const addToPackageScripts = async (cwd: string) => {
       {
         type: 'input',
         name: 'scriptName',
-        message: 'Enter name of script to add to package.json',
+        message: 'Enter the name of the script to add to package.json',
         default: scriptName,
       },
     ]);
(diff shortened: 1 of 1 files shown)


src/commands/init.js
```diff
@@ -146,6 +147,66 @@ async function init() {
...
+  );
+
+  const scripts = pjson.scripts || {};
+
+  const haulScript = Object.keys(scripts).find(name =>
+    scripts[name].includes('haul'));
+
+  if (haulScript) {
+    ora().info(
+      `Haul is already in your package's scripts. You can start it by running '${getRunScript(haulScript)}'`,
```
knowbody: do you mean in `package.json`? If so I'd just say:
Haul already exists in your package.json. Start Haul by running...

src/commands/init.js
```diff
@@ -146,6 +147,66 @@ async function init() {
...
+
+  if (
+    scripts.start &&
+    scripts.start !== 'node ./node_modules/react-native/local-cli/cli.js start'
+  ) {
+    const result = await inquirer.prompt([
+      {
+        type: 'input',
+        name: 'scriptName',
+        message: 'Enter name of script to add to package.json',
```
knowbody: I'd say:
> Enter the name of the script you want to add to package.json
or
> Enter the name of the script to add to package.json

src/commands/init.js
```diff
@@ -146,6 +147,66 @@ async function init() {
...
+  const progress = ora(
+    `Adding \`${scriptName}\` script to your package.json`,
+  ).start();
+
+  await sleep();
+
+  fs.writeFileSync(path.join(cwd, 'package.json'), JSON.stringify(pjson));
+
+  progress.succeed(
+    `You can now start Haul by running '${getRunScript(scriptName)}'`,
```
knowbody: > Start Haul by running...

src/commands/init.js
grabbou: Yeah, thanks!

src/commands/init.js
```diff
@@ -146,6 +147,66 @@ async function init() {
...
+ */
+const addToPackageScripts = async (cwd: string) => {
+  const pjson = JSON.parse(
+    fs.readFileSync(path.join(cwd, 'package.json')).toString(),
+  );
+
+  const scripts = pjson.scripts || {};
+
+  const haulScript = Object.keys(scripts).find(name =>
+    scripts[name].includes('haul'));
```
satya164: This will match words with the text `haul` in them. Probably better just to check `scripts[name] === 'haul start'`

src/commands/init.js
```diff
@@ -146,6 +147,66 @@ async function init() {
...
+
+  if (
+    scripts.start &&
+    scripts.start !== 'node ./node_modules/react-native/local-cli/cli.js start'
+  ) {
+    const result = await inquirer.prompt([
+      {
+        type: 'input',
+        name: 'scriptName',
+        message: 'Enter the name of the script to add to package.json',
```
satya164: Not fully clear what is script here, probably give an example - `e.g. - haul for yarn run haul` or something similar

src/commands/init.js
```diff
@@ -146,6 +147,66 @@ async function init() {
...
+        default: scriptName,
```
satya164: We should default to something else than `start`, I think `haul` is better, I think most of the time I'll go with this anyways

PR #102, opened 2017-04-05
Adds systrace middleware and removes logger boilerplate

grabbou:
Also, implements `logger.debug` used by loggerMiddleware and systraceMiddleware so that the output stays consistent.

### DIFF
--- a/src/server/middleware/systraceMiddleware.js
+++ b/src/server/middleware/systraceMiddleware.js
@@ -1,15 +1,16 @@
 /**
  * Copyright 2017-present, Callstack.
  * All rights reserved.
+ * 
+ * @flow
  */
+import type { Middleware, $Request } from 'express';
+
 const fs = require('fs');
 const dedent = require('dedent');
 const logger = require('../../logger');
 
-/**
- * Systrace middleware
- */
-function systraceMiddleware(req, res, next) {
+const systraceMiddleware: Middleware = (req: $Request, res, next) => {
   if (req.path !== '/systrace') {
     next();
     return;
@@ -17,6 +18,7 @@ function systraceMiddleware(req, res, next) {
 
   const path = `/tmp/react_native_${Date.now()}.json`;
 
+  // $FlowFixMe: rawBodyMiddleware adds `rawBody` to all requests
   fs.writeFileSync(path, req.rawBody);
 
   logger.debug('Systrace', `/tmp/react_native_${Date.now()}.json`);
@@ -32,6 +34,6 @@ function systraceMiddleware(req, res, next) {
     copy it.
   `,
   );
-}
+};
 
 module.exports = systraceMiddleware;
--- a/src/server/middleware/systraceMiddleware.js
+++ b/src/server/middleware/systraceMiddleware.js
@@ -9,13 +9,9 @@ import type { Middleware, $Request } from 'express';
 const fs = require('fs');
 const dedent = require('dedent');
 const logger = require('../../logger');
+const chalk = require('chalk');
 
 const systraceMiddleware: Middleware = (req: $Request, res, next) => {
-  if (req.path !== '/systrace') {
-    next();
-    return;
-  }
-
   const path = `/tmp/react_native_${Date.now()}.json`;
 
   fs.writeFile(path, req.rawBody, err => {
@@ -24,19 +20,18 @@ const systraceMiddleware: Middleware = (req: $Request, res, next) => {
       return;
     }
 
-    logger.debug('Systrace', `/tmp/react_native_${Date.now()}.json`);
-
-    res.end(
-      dedent`
-      Your trace report was saved at ${path}.
+    const message = dedent`
+      We've saved the trace report at ${chalk.bold(path)}
       
-      You can open it with Google Chrome by navigating to 'chrome://tracing'
-      and clicking 'load'.
+      You can open the following trace report in Google Chrome by navigating to
+      'chrome://tracing' and clicking 'load'.
+    `;
 
-      The path to the report was printed by Haul in your Terminal so you can
-      copy it.
-    `,
-    );
+    logger.log('\n');
+    logger.info(message);
+    logger.log('\n');
+
+    res.end(message);
   });
 };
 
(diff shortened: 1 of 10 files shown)


src/server/middleware/systraceMiddleware.js
```diff
@@ -0,0 +1,39 @@
...
+  logger.debug('Systrace', `/tmp/react_native_${Date.now()}.json`);
+
+  res.end(
+    dedent`
+    Your trace report was saved at ${path}.
+    
+    You can open it with Google Chrome by navigating to 'chrome://tracing'
+    and clicking 'load'.
+
+    The path to the report was printed by Haul in your Terminal so you can
```
satya164: This line feels unnecessary

src/server/middleware/systraceMiddleware.js
```diff
@@ -0,0 +1,39 @@
...
+ * @flow
+ */
+import type { Middleware, $Request } from 'express';
+
+const fs = require('fs');
+const dedent = require('dedent');
+const logger = require('../../logger');
+
+const systraceMiddleware: Middleware = (req: $Request, res, next) => {
+  if (req.path !== '/systrace') {
```
satya164: I think it'll be better to do things like `app.use('/systrace', systraceMiddleware)` than having these checks in each middleware

src/server/middleware/systraceMiddleware.js
```diff
@@ -0,0 +1,39 @@
...
+    next();
+    return;
+  }
+
+  const path = `/tmp/react_native_${Date.now()}.json`;
+
+  // $FlowFixMe: rawBodyMiddleware adds `rawBody` to all requests
+  fs.writeFileSync(path, req.rawBody);
```
satya164: Does it need to be sync? I think using async version will be pretty easy here.

We also need error handling.

src/server/middleware/systraceMiddleware.js
grabbou: Why? Packager already does that plus I don't think it's obvious to check terminal for additional output.

src/server/middleware/systraceMiddleware.js
grabbou: Not sure. From a developer point of view, it would make sense to use that pattern and abstract the path away. In case of RN-specific middleware, having this configurable doesn't make any sense given that it's hardcoded in React Native and should be tested in Haul to match these values.

src/server/middleware/systraceMiddleware.js
grabbou: Good point.

src/server/middleware/systraceMiddleware.js
satya164: It's not configurable. We're still hardcoding it where we define the server. Except we use express's feature instead of doing this manual check in every middleware.

src/server/middleware/systraceMiddleware.js
grabbou: The content of that method is displayed in an `Alert` inside iOS client - it cannot be copied. With Haul running in the background, I can imagine users missing that.

satya164: Yes, but you already print the path the line above. If user misses that line, why won't they miss this line?

This just seems silly

```
Here is the path of something /path/to/something

We have printed the path in the terminal
```

src/server/middleware/systraceMiddleware.js
```diff
@@ -0,0 +1,43 @@
...
+    if (err) {
+      next(err);
+      return;
+    }
+
+    logger.debug('Systrace', `/tmp/react_native_${Date.now()}.json`);
+
+    res.end(
+      dedent`
+      Your trace report was saved at ${path}.
```
satya164: Add `path` in `chalk.bold` for emphasis

src/server/middleware/systraceMiddleware.js
grabbou: screen shot 2017-04-05 at 15 24 11

Note `logger.debug` vs `res.end`

src/server/middleware/systraceMiddleware.js
```diff
@@ -0,0 +1,43 @@
...
+      
+      You can open it with Google Chrome by navigating to 'chrome://tracing'
+      and clicking 'load'.
+
+      The path to the report was printed by Haul in your Terminal so you can
```
satya164:
```
You can open the following trace report in Google Chrome by navigating to 'chrome://tracing' and clicking 'load'.

We've saved the trace report at ${chalk.bold(path)}
```

grabbou: Why to revert the order? The path information is more important

src/server/middleware/systraceMiddleware.js
satya164: Hm. Can we print the whole message in terminal too?

src/server/middleware/systraceMiddleware.js
grabbou: Chalk won't work for sending plain text to React Native, right?

src/server/middleware/systraceMiddleware.js
satya164: That's why it's at the end. It's easier to notice what's at the end in a long running server than what's in middle.

Oh, note that this is for printing in the terminal, not for alert message

src/server/middleware/systraceMiddleware.js
satya164: > Oh, note that this is for printing in the terminal, not for alert message

src/server/middleware/systraceMiddleware.js
grabbou: I think that given our logging system that highlights type of log, it makes sense to stay in the first line.

screen shot 2017-04-05 at 15 34 03

satya164: Okay

src/server/middleware/systraceMiddleware.js
```diff
@@ -0,0 +1,38 @@
...
+  fs.writeFile(path, req.rawBody, err => {
+    if (err) {
+      next(err);
+      return;
+    }
+
+    const message = dedent`
+      We've saved the trace report at ${chalk.bold(path)}
+      
+      You can open the following trace report in Google Chrome by navigating to
```
satya164: `You can open the trace report`

No `following` since it comes before the message

Issue #104, opened 2017-04-05
Feature Request: --platform windows


berwyn:
### Current Behavior
I use [React Native for Windows](https://github.com/Microsoft/react-native-windows) and would love if `--platform windows` was a thing I could use :)

### Expected Behavior
`--platform windows` runs and compiles code in my `.windows.tsx` files

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  0.2.1
| react-native     |  0.43.1
| node             | 7.7.1
| npm or yarn      | 0.21.3

acoates-ms:
#418 should allow you to use RN for windows.

PR #105, opened 2017-04-05
[WIP]: Add Windows Support

berwyn:
WORK IN PROGRESS -- PLEASE DON'T MERGE

As I mentioned in #104, I need support for Windows on my project in order to use Haul. I've made a first pass at it, but there's a few issues. Namely, it seems that the console polyfill fails on Windows and prevents the app from bootstrapping:

image

I'm not 100% sure how the console polyfill works, so I'd love some direction on where to take it from here!

### DIFF
--- a/src/commands/bundle.js
+++ b/src/commands/bundle.js
@@ -140,6 +140,10 @@ module.exports = {
           value: 'android',
           description: 'Builds Android bundle',
         },
+        {
+          value: 'windows',
+          description: 'Builds Windows bundle',
+        },
       ],
       example: 'haul bundle --platform ios',
     },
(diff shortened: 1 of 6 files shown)


berwyn:
Digging into this a bit more, it seems like there might be a race condition here between Chakra initialising its globals and the Haul-generated package calling initializeCore. Still digging to find which code is being problematic here, but I tend to think it's something with how Haul's `polyfillEnvironment` works, given that this works with the RN Packager but not Haul.

satya164:
cc @axemclion

grabbou:
CC: @rozele how are you integrating with packager in terms of setting up environment? Are there any pitfalls in particular that you would be interested in solving here?

grabbou:
Thanks for your contribution. I am going to close it for now due to inactivity and no real interest from Callstack to land it. If there's anyone willing to contribute to this, please get in touch.

PR #108, opened 2017-04-06
Fix iOS build phase

Kerumen:
In this commit https://github.com/callstack-io/haul/pull/90/commits/afd93e1b42c050b9690d81419464e5ceb819ffb2 @grabbou changed the comparison from `>` to `===`.
I don't know if it was intended but it broke the script. If a section is matched and replaced it doesn't save but exits with an error.

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -313,7 +313,7 @@ const addToXcodeBuild = async (cwd: string) => {
     );
   });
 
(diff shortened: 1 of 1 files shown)


grabbou:
Thanks, releasing patch version now. Some debugging slipped through 👎

grabbou:
Released as 0.3.1

PR #109, opened 2017-04-06
Fix small typo

baires:
Fix small typo on Readme


grabbou:
This is not a typo. That is needed to pass options to underlying command. Without it, you would pass platform option to yarn itself, not haul. 

When running haul directly without yarn that is not needed.

PR #127, opened 2017-04-18
Fail gracefully on source mapping lookup

ds300:
If input source mappings are incomplete, stack frame lookups can fail and return nully values. I have observed this when trying to set up Haul to load TypeScript files. This causes the symbolicate endpoint to fail with JS's peculiar version of a NPE.

image

It would be better to return the original stack frame for mappings that cannot be resolved. Then at least the mappings that *were* resolved can be seen. The underlying cause of the incomplete mappings is kind of beside the point. Better some than none, right?

This PR turns this:

image

into this:

image

Which, perhaps you'll agree, is a distinct improvement. Plus no nasty error messages in the webpack log.

### DIFF
--- a/src/server/middleware/symbolicateMiddleware.js
+++ b/src/server/middleware/symbolicateMiddleware.js
@@ -121,6 +121,14 @@ function create(compiler: *): Middleware {
           column: originalFrame.column,
         });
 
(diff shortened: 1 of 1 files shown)


grabbou:
Thank you both!

PR #131, opened 2017-04-19
add config path as an option

mathieudutour:
Allow to specify the path to the config file as an option

`haul start --config ./path/to/config.js`


satya164:
Looks reasonable. Can you check if both absolute and relative paths work?

cc @grabbou

mathieudutour:
oh no good point, only relative path will work with that

satya164:
Okies, should be easy to support both using `path`. Can you do the change?

mathieudutour:
done

PR #135, opened 2017-04-21
[WIP] Add initial support for HMR

migueloller:
This adds initial support for HMR but is still not working.

Specifically, calls to `module.hot.check` time out when retrieving the manifest and `webpack-hot-middleware` will require an additional polyfill for `window.EventSource` to satisfy [this](https://github.com/glenjamin/webpack-hot-middleware/blob/40745357fa0c069204d01c993371bdbcd2e6fc74/client.js#L35-L45) so that it works without having remote JS debugging enabled.

Would love to hear some feedback on this.

cc: @satya164

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "strip-ansi": "^3.0.1",
     "webpack": "^2.3.1",
     "webpack-dev-middleware": "^1.10.1",
+    "webpack-hot-middleware": "^2.18.0",
     "ws": "^2.2.2"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


migueloller:
Adding `publicPath: 'http://localhost:8081/',` to the Webpack config `output` solves the issue with the timeout but results in the following error:

image

satya164:
Ahh.. I think it's because the webpack loads scripts with script tags. We need to see how React Native's packager does it. Probably just uses eval.

cc @grabbou

migueloller:
@satya164,

Yeah, it uses JSONP as you can see [here](https://github.com/webpack/webpack/blob/54aa3cd0d6167943713491fd5e1110b777336be6/lib/JsonpMainTemplate.runtime.js#L12-L19)

grabbou:
Thanks @migueloller for your input. It's been a great source of learning for us and we have came up with a PR #182 that we are hoping to get merged soon :)

PR #139, opened 2017-04-24
Opens an editor when tapping a stack frame.

### LINKED ISSUE #123: Open File in Editor Support
In the React Native simulator, you can tap a stack frame and your editor will open to that file.

[To do this](https://github.com/facebook/react-native/blob/master/Libraries/Core/Devtools/openFileInEditor.js#L17), they POST to '/open-stack-frame' passing along `{file, lineNumber}`.

What happens next is [kinda involved](https://github.com/facebook/react-native/blob/master/local-cli/server/util/launchEditor.js).

Supporting every single editor doesn't mentally scale, but maybe just offering a callback/plugin to haul where one could define their editor?

Would you guys be open to a PR supporting a dumbed down version of this to start?

skellock:
This PR adds support for tapping on a stack trace in the simulator.  It will open the user's editor to the right line/column.

There's a few ways to use it, here's the easiest:

```sh
export REACT_EDITOR=code
```

This works too:

```sh
export REACT_EDITOR=/usr/local/bin/code
```

[The library we're using](https://github.com/lahmatiy/open-in-editor) supports about a dozen editors and it knows where to look for the standard installs.  It knows how to launch them and use the right order for parameters.

For custom installs, we had to invent some things:

```sh
export HAUL_EDITOR=sublime
export HAUL_EDITOR_CMD="/path/to/Sublime Super Mega Beta.app"
export HAUL_EDITOR_PATTERN="-r -g {filename}:{line}:{column}"
```

That's for more advanced setups.

Out of the box, if no environment variables are found, it prints some instructions:

image

I've tried atom, code, vim, and sublime on MacOS 10.11.6.  I haven't tried on Windows or Linux, however, the library we're using supports those.  I will give it a test tomorrow on both and sound the air raid horns if something is awry.

Vim opens up in Terminal, not iTerm.  This makes me sad in the pants.  I don't know how to fix.

Feedback welcome. (e.g. should we add a command line option too?).

Fixes #123

### DIFF
--- a/src/server/middleware/openInEditorMiddleware.js
+++ b/src/server/middleware/openInEditorMiddleware.js
@@ -1,0 +1,148 @@
+/**
+ * Copyright 2017-present, Callstack.
+ * All rights reserved.
+ *
+ * @flow
+ *
+ * --- OVERVIEW ---
+ *
+ * React Native apps show a stack trace when an error occurs. Each
+ * frame can be tapped on by the user. When they do, React Native will
+ * send a POST to /open-stack-frame passing along {file, lineNumber}.
+ *
+ * Haul should then open the user's editor to the file and lineNumber.
+ *
+ * What makes this gloriously fragile is, people use different editors
+ * on different platforms.
+ *
+ * We will rely on the NPM package `open-in-editor` to provide this
+ * functionality. (read: scapegoat).
+ */
+import type { $Request, $Response, Middleware } from 'express';
+import type { ReactNativeStackFrame } from '../../types';
+
+// A configuration object fed into `open-in-editor`.
+type OpenInEditorConfig = {
+  editor?: string | null,
+  cmd?: string | null,
+  pattern?: string | null,
+};
+
+const basename = require('path').basename;
+const openInEditor = require('open-in-editor');
+const editors = Object.keys(require('open-in-editor/lib/editors'));
+const messages = require('../../messages');
+const logger = require('../../logger');
+
+// does `open-in-editor` know how to detect this editor?
+const isKnownEditor = (editor: string): boolean =>
+  editors.indexOf(basename(editor)) >= 0;
+
+/**
+ * Detect the editor type to use. We use just the basename in case the user
+ * has passed a full path. Just the name work here though. Examples:
+ *
+ *   - sublime
+ *   - atom
+ *   - /usr/local/bin/atom
+ *
+ */
+function detectEditorType(): string | null {
+  // check for HAUL_EDITOR first
+  if (process.env.HAUL_EDITOR) {
+    const editor = basename(process.env.HAUL_EDITOR);
+    if (isKnownEditor(editor)) return editor;
+  }
+
+  // fallback to REACT_EDITOR
+  if (process.env.REACT_EDITOR) {
+    const editor = basename(process.env.REACT_EDITOR);
+    if (isKnownEditor(editor)) return editor;
+  }
+
+  return null;
+}
+
+/**
+ * If provided, this will be the editor command that we shell out to. You only
+ * need to use this if you've got a specific script or symlink you're trying
+ * to target.
+ */
+function detectEditorCmd(): string | null {
+  // the specific haul command gets precident
+  if (process.env.HAUL_EDITOR_CMD) return process.env.HAUL_EDITOR_CMD;
+
+  // use HAUL_EDITOR unless it's a shortcut (like just the word 'atom')
+  if (process.env.HAUL_EDITOR && !isKnownEditor(process.env.HAUL_EDITOR)) {
+    return process.env.HAUL_EDITOR;
+  }
+
+  // fallback to REACT_EDITOR
+  if (process.env.REACT_EDITOR && !isKnownEditor(process.env.REACT_EDITOR)) {
+    return process.env.REACT_EDITOR;
+  }
+
+  return null;
+}
+
+/**
+ * If provided, this allows people to control how the arguments are passed
+ * to their editor.  For example: -r -g {filename}:{line}:{column}
+ */
+function detectEditorPattern(): string | null {
+  return process.env.HAUL_EDITOR_PATTERN || null;
+}
+
+// configure the editor open library
+const editorConfig: OpenInEditorConfig = {};
+editorConfig.editor = detectEditorType();
+editorConfig.cmd = detectEditorCmd();
+
+// null is a valid pattern, so we need to only add it if we mean it
+const pattern = detectEditorPattern();
+if (pattern) {
+  editorConfig.pattern = pattern;
+}
+
+const hasValidConfig = editorConfig.editor || editorConfig.cmd;
+
+// create the middleware
+function create(): Middleware {
+  // create an editor we'll use to open editor.
+  const opener = openInEditor.configure(editorConfig);
+
+  /**
+   * The Express middleware for opening in the editor.
+   */
+  function openInEditorMiddleware(req: $Request, res: $Response, next) {
+    // only allow the appropriate path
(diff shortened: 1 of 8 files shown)


src/server/middleware/openInEditorMiddleware.js
```diff
@@ -0,0 +1,148 @@
...
+ * has passed a full path. Just the name work here though. Examples:
+ *
+ *   - sublime
+ *   - atom
+ *   - /usr/local/bin/atom
+ *
+ */
+function detectEditorType(): string | null {
+  // check for HAUL_EDITOR first
+  if (process.env.HAUL_EDITOR) {
```
grabbou: Instead of using env variables, how about we leverage webpack.haul.js? I think we need to change its name, but that can be done later in next PR?

grabbou: I think if we put it into a Javascript file, all we need is:
```
exports.editor = string | (fileInfo) => void
```
that when:
- given a string, is a path to executable, e.g. `code`
- when given a function, it receives file path, line number and is supposed to return a command that will get executed, e.g. `code -r -g ./abc.js`

skellock: Unfortunately with that approach, everyone would be forced to use the same editor. At our shoppe, I'd be hung for treason. ☠️

satya164: yeah, reading env variable for editor seems reasonable.

do we need a separate `HAUL_EDITOR` though? Won't you use the same one you use for React anyway?

skellock: I think 9 out of 10 devs would be fine with REACT_EDITOR, but the 10th person who uses `nvim`, a custom symlink, or another editor unknown by `open-in-editor` can still make this work. I didn't want to exclude them. 

I can remove it if you feel its too complicated.

satya164: We can definitely support the following too I think, if supporting crazy editors is a concern

```
export REACT_EDITOR_CMD="/path/to/Sublime Super Mega Beta.app"
export REACT_EDITOR_PATTERN="-r -g {filename}:{line}:{column}"
```

skellock: I'm down with renaming HAUL_* to REACT_* if you'd like.

skellock: I'm also down with deleting that code until a real human asks for it. I kinda just guessed people might need it.

satya164: I think that's a good option too. cc @grabbou

src/server/middleware/openInEditorMiddleware.js
```diff
@@ -0,0 +1,148 @@
...
+}
+
+/**
+ * If provided, this will be the editor command that we shell out to. You only
+ * need to use this if you've got a specific script or symlink you're trying
+ * to target.
+ */
+function detectEditorCmd(): string | null {
+  // the specific haul command gets precident
+  if (process.env.HAUL_EDITOR_CMD) return process.env.HAUL_EDITOR_CMD;
```
grabbou: Do we need `HAUL_EDITOR_CMD`? I think `HAUL_EDITOR` being either known editor or an absolute path is perfectly fine?

src/server/middleware/openInEditorMiddleware.js
```diff
@@ -0,0 +1,148 @@
...
+
+  // use HAUL_EDITOR unless it's a shortcut (like just the word 'atom')
+  if (process.env.HAUL_EDITOR && !isKnownEditor(process.env.HAUL_EDITOR)) {
```
grabbou: Also, I think we should rename that to `REACT_EDITOR`. If someone wants to use something specific for `Haul` projects, he can pass the variable inline?

src/server/middleware/openInEditorMiddleware.js
grabbou: > I'm down with renaming HAUL_* to REACT_* if you'd like.

Just left a comment, I think it's good to do it.

grabbou:
Okay, I am fine with this PR. Just two things I outlined in my review (mostly `nit`s)

src/server/middleware/openInEditorMiddleware.js
```diff
@@ -0,0 +1,148 @@
...
+ * functionality. (read: scapegoat).
+ */
+import type { $Request, $Response, Middleware } from 'express';
+import type { ReactNativeStackFrame } from '../../types';
+
+// A configuration object fed into `open-in-editor`.
+type OpenInEditorConfig = {
+  editor?: string | null,
+  cmd?: string | null,
+  pattern?: string | null,
```
grabbou: Wouldn't `pattern?: ?string` be the same?

src/server/middleware/openInEditorMiddleware.js
```diff
@@ -0,0 +1,148 @@
...
+
+const basename = require('path').basename;
+const openInEditor = require('open-in-editor');
+const editors = Object.keys(require('open-in-editor/lib/editors'));
+const messages = require('../../messages');
+const logger = require('../../logger');
+
+// does `open-in-editor` know how to detect this editor?
+const isKnownEditor = (editor: string): boolean =>
+  editors.indexOf(basename(editor)) >= 0;
```
grabbou: nit: we can use `includes`

src/server/middleware/openInEditorMiddleware.js
```diff
@@ -0,0 +1,148 @@
...
+
+  return null;
+}
+
+/**
+ * If provided, this will be the editor command that we shell out to. You only
+ * need to use this if you've got a specific script or symlink you're trying
+ * to target.
+ */
+function detectEditorCmd(): string | null {
```
grabbou: nit: we can use `?string` AFAIK

skellock:
Sweet. Ya, I saw you guys are super busy (woot beta 9!). This PR is a nice-to-have.

So I'll get those nits addressed. I'm still a noob with Flow, so tbh my goal is to shut the checker up, not to actually write code that makes sense. 🎢 🤢 

The `HAUL_EDITOR_CMD` is a thing to address the case of some naming their editor command something that `open-in-editor` doesn't recognize. `nvim` for example. It's also required if we want to have a fallback for "other" editor.

We could ditch all 3 `HAUL_*` env vars if you want. We'd just lose flexibility in opening certain editors. Which I'm 100% fine with as it won't affect me! :shipit: 

Another option is we keep it, but remove it from the giant cli message (keeping only our preferred approach of `REACT_EDITOR`) and I add a new doc called `tips.md` and with this edge.

So keep, kill, or demote, I'm cool with all 3. Lemme know.

satya164:
I think we can rename `HAUL_*` to `REACT_*` for now and we should be good to go

src/server/middleware/openInEditorMiddleware.js
satya164: ya, `?string` is good

skellock:
open in editor demo

grabbou:
side question: what did you use to capture the gif?

satya164:
I use https://getkap.co/

skellock:
[ScreenFlow](https://www.telestream.net/screenflow/overview.htm).  Pricey, but worth it every Canadian peso.

I don't use most of the features, but, the timeline editor is 💯 .

* record me flailing around
* crop the viewport
* set some in & out points by clicking
* slice up & delete some dead air
* double-click a slice and speed up some of them to 400%
* export as gif

Kap (and kin) is great if you're a one-take no-mistake kinda fellow.  And that is decidedly not me.

Cheers for the merge!

PR #145, opened 2017-05-03
add `--platform all` bundle option

mathieudutour:


docs/CLI Commands.md
```diff
@@ -42,9 +42,9 @@ Example: `haul bundle --platform android --bundle-output build/index.bundle --as
 
 You can specify following parameters to configure the bundle generation:
 
-### `--platform <ios|android>`
+### `--platform <ios|android|all>`
 
-The platform to build the bundle for.
+The platform to build the bundle for. `all` will build the bundle for both `android` and `ios` and is slower.
```
satya164: I feel like it'll be future-proof to do `platform=ios,android` rather than `platform=all`. Will be useful when we add more platfroms such as windows, web etc. Might wanna deprecate `platform=all` cc @grabbou

mathieudutour:
closing this PR, will open another one from a branch (not my master) since I have some other fixes that I need

PR #147, opened 2017-05-05
add --platform all bundle option

mathieudutour:
see #145


grabbou:
I think ultimately we want to deprecate `platform=all` behaviour and prefer comma separated list instead. 

Would you mind submitting such pull request once again?

Thanks for your effort, but I am going to close this one for now

PR #148, opened 2017-05-06
check for global.self first

spypsy:
This is actually running in the browser and throwing an error that can make script injection fail


satya164:
Thank you for the PR

PR #161, opened 2017-06-08
Remove package.json script

grabbou:
Default name of package.json script is `haul` and defaults to `haul start`. Having that value set, it makes it hard for others to run other commands, since now `yarn haul` executes such script.

By removing that phase and package.json script entirely, users can just run `yarn haul start` or any other option. 

Also, it eliminates confusion since in docs, we recommend executing through `yarn`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ project.ext.react = [
 To configure your project to use haul, run the following:
 
 ```bash
-yarn run haul init
+yarn haul init
 ```
 
 This will automatically add the configuration needed to make Haul work with your app, e.g. add `webpack.haul.js` to your project, which you can customise to add more functionality.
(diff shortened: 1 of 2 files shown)


grabbou:
CC: @skellock

satya164:
This can be resolved by just renaming the default script name to `serve` or something else IMO.

Even if we document yarn (we should document both npm and yarn), there's no dependency on yarn. This changes makes it harder to get started if someone is using npm.

grabbou:
Question if it's better to have `npm run server` opposed to `npm run haul start` or `yarn haul start`.

satya164:
Note that default is `npm start`/`yarn start` in most RN apps. `yarn run serve` or `npm run serve` is just a fallback and not the preferred script.

Issue #157, opened 2017-05-26
Haul does not recognize react-native as a peerDependency


kcjonson:
### Your Environment
We work in a monorepo that hoists common dependencies up to the root folder with Lerna. This means that `react-native` itself is not installed in the project folder, but is two levels higher in the filesystem.

### Current Behavior
`haul init` throws an error that it can't find react-native
`haul start --platform ios` also cannot find react-native and exits

### Expected Behavior
Haul should resolve up the file system tree as node/webpack does to find a version of react-native

### Haul Configuration (webpack.haul.js)
Minimal config in a test application

| software         | version
| ---------------- | -------
| Haul             |   v0.6.0
| react-native     |  44.0
| node             |  v6.9.1
| npm or yarn      | npm 3.10.8

kcjonson:
A couple ways to fix this up:
1) Use `node-resolve` (or similar) to resolve to a react-native install (package [here](https://github.com/substack/node-resolve))
2) Add a `react-native-location` field to the config file
2.1) `haul init` prompts the user to enter the location of react-native then writes it to the config

I prefer option one. Any foreseen issues with going in that direction?

grabbou:
Shouldn't require work anyway even if React Native is installed two levels up the directory? All in all, that's how require traverses up the file system. 

Have you installed haul globally or locally? Have you tried installing it on the same level as React Native? Does it work then?

kcjonson:
Two things that I found

**src/commands/init.js**: `if (pak.dependencies['react-native']) {` won't allow `peerDependencies` but thats a pretty easy fix to swap to `if (pak.dependencies['react-native'] || pak.peerDependencies['react-native']) {` 

then the only other blocker is actually in building the haste map

**/src/utils/makeReactNativeConfig.js** has the location of `react-native` hard coded to `directories: [path.resolve(root, 'node_modules/react-native')],` which may not actually be its final install location. In the case of the repo I'm looking at it ends up being `../../node_modules/react-native` but that will vary by how deeply the project is nested in your repo.  The idea would be to use `node-resolve` or something to find where it really is when that config is built rather than use a relative path.

kcjonson:
to add a bit more detail, this is essentially how the directory structure ends up after a `lerna bootstrap`

```
main/
   mobile/
      appname/
         webpack.haul.js
         package.json
         index.ios.js
   node_modules/
      haul-cli/
      react/
      react_native/


```

satya164:
> The idea would be to use node-resolve

[...]

PR #163, opened 2017-06-08
Check in all places for react-native

grabbou:
Issue #157 

This PR checks in `dev` and `peer` just in case someone had it set up differently.


grabbou:
The issue can be closed after we implement `node-resolve` which @Krizzu is working on I believe.

PR #172, opened 2017-06-14
Fix asset size when bundling larger asset

satya164:


grabbou:
Ready for review? (I know not, but pinging xd)

grabbou:
LGTM, thanks!

Will test it before releasing on master by running in the project I was facing the issue.

PR #178, opened 2017-06-26
Better feedback for compilation errors

Krizzu:
Custom reporter for webpack middleware to check if there were any `compilation errors`. If yes,  show it to a user.


Krizzu:
A wild gif appears!

Super gif

Krizzu:
This functionality already exists in Hual. If you don't have it, clean `cache` and reinstall `node_modules`

seklyza:
Are you sure it exists?

For example, I am using TypeScript and I would like to have the red error screen (or something like that) in the device/AVD is there's a compilation/linting error with webpack. And yes, I do have webpack.NoEmitOnErrorsPlugin() in the config.

PR #181, opened 2017-06-27
Fix assetLoader not using platform path separator

JulioC:
Fixes https://github.com/callstack-io/haul/issues/180

### DIFF
--- a/src/loaders/assetLoader.js
+++ b/src/loaders/assetLoader.js
@@ -36,6 +36,7 @@ module.exports = async function assetLoader() {
     // Asset is not an image
   }
 
+  const pathSepPattern = new RegExp(`\\${path.sep}`, 'g');
   const filepath = this.resourcePath;
   const dirname = path.dirname(filepath);
   const url = path.relative(config.root, dirname);
@@ -43,7 +44,7 @@ module.exports = async function assetLoader() {
   const assets = path.join('assets', config.bundle ? '' : config.platform);
   const suffix = `(@\\d+(\\.\\d+)?x)?(\\.(${config.platform}|native))?\\.${type}$`;
   const filename = path.basename(filepath).replace(new RegExp(suffix), '');
-  const longname = `${`${url.replace(/\//g, '_')}_${filename}`
+  const longname = `${`${url.replace(pathSepPattern, '_')}_${filename}`
     .toLowerCase()
     .replace(/[^a-z0-9_]/g, '')}.${type}`;
 
@@ -129,7 +130,7 @@ module.exports = async function assetLoader() {
     this.emitFile(dest, item.content);
   });
 
-  let publicPath = `__webpack_public_path__ + ${JSON.stringify(path.join('/', assets, url))}`;
+  let publicPath = `__webpack_public_path__ + ${JSON.stringify(path.join('/', assets, url).replace(pathSepPattern, '/'))}`;
 
   if (config.publicPath) {
     // support functions as publicPath to generate them dynamically


src/loaders/assetLoader.js
```diff
@@ -129,7 +130,7 @@ module.exports = async function assetLoader() {
     this.emitFile(dest, item.content);
   });
 
-  let publicPath = `__webpack_public_path__ + ${JSON.stringify(path.join('/', assets, url))}`;
+  let publicPath = `__webpack_public_path__ + ${JSON.stringify(path.join('/', assets, url).replace(pathSepPattern, '/'))}`;
```
satya164: Does `__webpack_public_path__ + ${JSON.stringify(path.join(path.sep, assets, url))}` work here?

src/loaders/assetLoader.js
```diff
@@ -36,14 +36,15 @@ module.exports = async function assetLoader() {
     // Asset is not an image
   }
 
+  const pathSepPattern = new RegExp(`\\${path.sep}`, 'g');
```
satya164: Does this work on Unix?

src/loaders/assetLoader.js
JulioC: I don't think so. The first param would be just part of the path.

satya164: yes, but the other stuff are returned by webpack, so they should have proper path separators.

src/loaders/assetLoader.js
JulioC: It would, but I haven't tested it yet.

If `path.sep === '/'`, the generated RegExp would be `/\//g`.

src/loaders/assetLoader.js
JulioC: I don't get your point. The `path.join` would still use the platform specific `path.sep`, but react-native code expects it to be / (since it's from a URL), see: https://github.com/facebook/react-native/blob/1fa95ed390d96b4a54702acdf89594af1785a3fd/local-cli/bundle/assetPathUtils.js#L38

satya164: > /local-cli/bundle/assetPathUtils.js#L38

That code is not relevant since it's part of RN packager and not haul.

Can you add a comment explaining it? It's not immediately obvious why we need to do it.

satya164:
Can you rebase the PR?

src/loaders/assetLoader.js
JulioC: That code is used in the react-native runtime, on [/Libraries/Image/AssetSourceResolver.js#L27](https://github.com/facebook/react-native/blob/1fa95ed390d96b4a54702acdf89594af1785a3fd/Libraries/Image/AssetSourceResolver.js#L27), to resolve the asset filename ([/Libraries/Image/resolveAssetSource.js](https://github.com/facebook/react-native/blob/1fa95ed390d96b4a54702acdf89594af1785a3fd/Libraries/Image/resolveAssetSource.js))

JulioC:
@satya164 I'm not sure if I understand how I'd do it.

src/loaders/assetLoader.js
satya164: Interesting. Anyway, can you add a comment?

satya164:
@JulioC 

```
git checkout master
git pull upstream master master
git checkout patch-1
git rebase master
git push -f
```

JulioC:
Would it be possible for you to do that, using [this GitHub feature](https://github.com/blog/2243-rebase-and-merge-pull-requests)? If not, I can try doing it manually.

JulioC:
@satya164 Is there anything pending on this PR?

PR #183, opened 2017-06-29
Better Error messages

Krizzu:
Before:
screen shot 2017-06-29 at 15 31 00

After:
screen shot 2017-06-29 at 15 31 39

Errors are visible now, but stack trace is pointing back to `worker`.
Still WIP.
- [X] Better Error messages
- [x] Proper stack trace


Krizzu:
Stack trace is shown now.

screen shot 2017-06-30 at 12 09 42


This fix is only for remote debugging.

Issue #176, opened 2017-06-20
symbolicateMiddleware fails when platform is set to all


markbiddlecom:
### Current Behavior
When running `haul start --platform all`, `symbolicateMiddleware` throws whenever a log message is inspected.

```
TypeError: Cannot read property 'output' of undefined
    at createSourceMapConsumer (<redacted>/node_modules/haul-cli/src/server/middleware/symbolicateMiddleware.js:41:30)
    at symbolicateMiddleware (<redacted>/node_modules/haul-cli/src/server/middleware/symbolicateMiddleware.js:97:22)
    at Layer.handle [as handle_request] (<redacted>/node_modules/haul-cli/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (<redacted>/node_modules/haul-cli/node_modules/express/lib/router/index.js:317:13)
    at <redacted>/node_modules/haul-cli/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (<redacted>/node_modules/haul-cli/node_modules/express/lib/router/index.js:335:12)
    at next (<redacted>/node_modules/haul-cli/node_modules/express/lib/router/index.js:275:10)
    at statusPageMiddleware (<redacted>/node_modules/haul-cli/src/server/middleware/statusPageMiddleware.js:16:3)
    at Layer.handle [as handle_request] (<redacted>/node_modules/haul-cli/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (<redacted>/node_modules/haul-cli/node_modules/express/lib/router/index.js:317:13)
```

### Expected Behavior
No errors in the console. Either fail gracefully or (ideally) understand which compiler the source map should be loaded from.

The issue appears to be that when platform is `all`, the `compiler` passed to the middleware function is a `MultiCompiler` and contains an array of compilers: one for each platform. This may also be a problem with the other middleware components; I didn't check.

### Haul Configuration (webpack.haul.js)

`webpack.haul.js`
```js
const path = require('path');

module.exports = ({ platform }, defaults) => ({
  entry: `./src/index.${platform}.js`,
  module: {
    ...defaults.module,
    rules: [
      ...defaults.module.rules,
      {
        test: /\.(css|less)$/,
        loaders: [path.join(__dirname, "native-css-loader.js"), 'less-loader', '<redacted>=' + __dirname + '/less-preamble.json'],
      }
    ],
  },
  resolve: {
    ...defaults.resolve,
    modules: [
      "node_modules",
[...]

PR #185, opened 2017-07-03
Adds error symbolication in multi-platform mode.

skellock:
### Overview

A fix for #176 which revealed that symbolication didn't work while in `--platform all` mode.

### Implementation Concerns

So, I had thought that React Native passed the platform when did it's symbolicate request.  It doesn't.  😿 

What I'm doing now is "detecting" it by looking at the first frame's unresolved filename.  Somewhere in there it'll say `...index.ios.bundle?platform=ios&...`.  I'm matching the platform found there against the webpack `options.output.filename` config.  

Not award-winning, but it works.  Would be nicer if there was a easier way to get the platform from the config.  I created a helper `getCompilerByPlatform` where we can put that.

### Sneaky Fallback

Something presumptuous I snuck in (I know @grabbou & @satya164 will rightfully call me out on this) is, if we can't find the platform (due to an error in the calling middleware), we fallback to the first compiler.  This will generally be ok unless you're using platform-specific code.

I can remove this if y'all aren't onboard.

### Other Potential Problems

@markbiddlecom mentioned this `MultiCompiler` issue might be in other middleware too.  He's right.  [This might be](https://github.com/callstack-io/haul/blob/master/src/server/middleware/liveReloadMiddleware.js#L32) an issue as well.

### Housekeeping

This may collide with with PR #183 so if you wanna merge that first, ping me and I'll rebase this.

### DIFF
--- a/src/messages/index.js
+++ b/src/messages/index.js
@@ -32,4 +32,5 @@ module.exports = {
   gitAlreadyAdded: require('./gitAlreadyAdded'),
   gitNotFound: require('./gitNotFound'),
   editorNotConfigured: require('./editorNotConfigured'),
+  webpackCompilerPlatformNotFound: require('./webpackCompilerPlatformNotFound'),
 };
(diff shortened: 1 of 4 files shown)


grabbou:
All tested and works well. I think the fallback to first compiler is reasonable. The overall logic is 👍 .

PR #192, opened 2017-07-22
Update SCRIPT_SRC to work with rn@0.46.4

haohcraft:
Otherwise, it would throw errors:
```
../node_modules/haul/src/utils/haul-integrate.sh: line 5: cd: ../node_modules/haul/src/utils/../../../react-native/packager: No such file or directory

sed: /react-native-xcode.sh: No such file or directory

../node_modules/haul/src/utils/haul-integrate.sh: line 12: /packager.sh: Permission denied


```

### DIFF
--- a/src/utils/haul-integrate.sh
+++ b/src/utils/haul-integrate.sh
@@ -2,7 +2,7 @@
 
 THIS_DIR=$(dirname $0)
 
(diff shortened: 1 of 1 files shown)


grabbou:
Hey, looks like this is already fixed in https://github.com/callstack-io/haul/commit/870753d1df5f2d6f878584adb95a370bc2b5b169. Make sure you are using latest version :)

PR #197, opened 2017-07-27
Simplify eslint config with callstack-io config

thymikee:
**Summary**

* bump ESLint to v4
* defer configuration to `eslint-config-callstack-io`


src/loaders/assetLoader.js
```diff
@@ -137,9 +140,9 @@ module.exports = async function assetLoader() {
     );
   }
 
-  const hashes = pairs.map(item => hasha(item.content, { algorithm: 'md5' }));
+  const hashes = pairs.map(item => hasha(item.content, {algorithm: 'md5'}));
 
-  let info: ?{ width: number, height: number, type: string };
```
grabbou: Can we revert this in upstream `callstack` preset? I think spaces in between `{` are already widely used across our internal projects 💯

thymikee: ok! Just thought we go full `fb` config.

thymikee:
I've set up circle to run only tests on Node 4 and tests + eslint on Node 6. 
Because we install deps with `yarn`, there's no peerDeps problem npm v2 introduces.

.circleci/config.yml
```diff
@@ -0,0 +1,35 @@
...
+    working_directory: ~/haul
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+          - v1-dependencies-{{ checksum "package.json" }}
+          - v1-dependencies
+      - run:
+          name: update-yarn
+          command: sudo npm i -g yarn
```
satya164: Install `yarn` with

```
  - curl -o- -L https://yarnpkg.com/install.sh | bash
  - export PATH=$HOME/.yarn/bin:$PATH
```

grabbou: You can also use https://github.com/callstack-io/new-website/blob/master/.circleci/config.yml#L5 since that Docker build already has it (taken from react-navigation and new-website repositories)

grabbou: @thymikee including, as the referenced comment was already fixed.

grabbou:
I remember we have talked about this previously, but what is the motivation behind setting `4` and `6` versions of Node? 

Just to put more context here: We wanted to stick with the minimum version required by React Native.

.circleci/config.yml
grabbou: An interesting idea would be to create a custom Docker image. The one I referenced is cool, but is nowhere documented (no repo that I could find). Might be nice to have it for future projects as well (will simplify infrastructure a lot)

thymikee:
Reverted the Circle work because it's not in the scope of this PR
See: https://github.com/callstack-io/haul/pull/198

thymikee:
CI green 🎉

grabbou:
Is it possible we move all the custom rules to `callstack-io` preset so that we use the same company-wise config everywhere?

thymikee:
I'll adjust it tomorrow 👍

PR #198, opened 2017-07-27
Setup Circle with workflows for Node 4 and 6

thymikee:
cc @grabbou @satya164


grabbou:
Let's see if it passes. From my perspective, it would be great if it had more comments and spacing, so it's easier for people to analyse :)

thymikee:
Will do once it pass xD

PR #200, opened 2017-07-28
Integration testing suite for CLI

thymikee:
**Summary**

This is a work in progress PR, where I'll be adding integration tests to Haul CLI. 
Will send followup PRs with more tests (unit included) after this is done.

TODOs:
- [x] setup framework for integration testing
- [x] add basic integration tests for commands


thymikee:
I cannot yet figure out what's wrong with yarn install inside fixtures. It works locally. Also similar setup works fine inside Jest repo (also on circle 2, there's PR open).

Cc @bestander, maybe you have some smart thoughts?

bestander:
@thymikee, this does ring a bell.

1. Try latest Yarn 0.28.4, it has heaps of bugs fixed

2. Any chance that you run `yarn install` in multiple threads?
If so use `--mutex network` argument to make sure installs don't break global cache

thymikee:
@bestander thanks heaps! The `--mutex network` works just right 👌. I thought tests run serially, but I ran `npm run jest -i` instead of `npm run jest -- -i` . 
Thanks to `--mutex` I can investigate how the tests perform in parallel on Circle

thymikee:
@grabbou I decided to skip testing Node 4 for now, as it's super counterproductive for me because of the Haul buggy behaviour there. Can we, for now, stop requiring `test-node-4`?

thymikee:
There's still a lot of things to test, but at least we have a base now. I'll start adding some smaller unit ones in a followup to this

integration_tests/runHaul.js
```diff
@@ -0,0 +1,65 @@
+/**
```
thymikee: This file and files below are relevant

integration_tests/__tests__/start.test.js
```diff
@@ -0,0 +1,73 @@
...
+      expect(stdout).toMatch('Haul is now bundling your React Native app');
+      expect(stdout).toMatch('A fresh build may take longer than usual');
+      expect(stdout).toMatch(
+        `You can now run the app on your ${platform} device`,
+      );
+      done();
+      haul.kill();
+    }
+  });
+}
```
thymikee: This file and files above are relevant

grabbou:
I think we can disable `node 4` tests for now - let's make an issue so that we can keep track of supporting them again.

integration_tests/__tests__/init.test.js
```diff
@@ -0,0 +1,44 @@
...
+  haul.stdout.on('data', () => {
+    haul.stdin.write(ENTER_KEY);
+  });
+
+  haul.stdout.on('close', () => {
+    try {
+      const haulConfig = fs.readFileSync(CONFIG_FILE_PATH, 'utf8');
+      expect(haulConfig).toEqual(
+        // eslint-disable-next-line
+        'module.exports = ({ platform }) => ({\n  entry: `./index.${platform}.js`,\n});',
```
grabbou: How about using snapshot for this?

integration_tests/__tests__/start.test.js
```diff
@@ -0,0 +1,73 @@
...
+const stripAnsi = require('strip-ansi');
+
+const TEMP_DIR = path.resolve(os.tmpdir(), 'start_test');
+const TEST_PROJECT_DIR = path.resolve(
+  __dirname,
+  '../fixtures/react-native-with-haul',
+);
+
+beforeEach(() => {
+  run('yarn --mutex network', TEST_PROJECT_DIR);
```
grabbou: Why do we want to run `yarn` before each of the suites? `start` doesn't affect `node_modules` at all?

thymikee: I should switch it to beforeEach. You sure about that? I remember having troubles running it without react native available, but may revisit that

grabbou:
Overall it looks good - does anyone else have any ideas / objections / suggestions for this pull request? 

@thymikee I think it's best if we pair on Slack for few seconds and I'll test what we got here before hitting final merge. What do you think?

integration_tests/__tests__/start.test.js
grabbou: That would be weird, but worth double checking

thymikee: I mean beforeAll

thymikee:
Addressed the feedback.

integration_tests/__tests__/bundle.test.js
```diff
@@ -0,0 +1,48 @@
...
+const fs = require('fs');
+const { cleanup, run } = require('../utils');
+
+const TEST_PROJECT_DIR = path.resolve(
+  __dirname,
+  '../fixtures/react-native-with-haul',
+);
+
+beforeEach(() => {
+  run('yarn --mutex network', TEST_PROJECT_DIR);
```
grabbou: Here I think it should be `beforeAll` as well

integration_tests/__tests__/start.test.js
```diff
@@ -0,0 +1,71 @@
...
+    if (message.match(/ERROR/g)) {
+      done.fail(message);
+      haul.kill();
+    }
+
+    if (message.match('Built successfully!')) {
+      const stdout = messageBuffer.join('\n');
+      expect(stdout).toMatch('INFO  Ready at http://localhost:8081');
+      expect(stdout).toMatch('Haul is now bundling your React Native app');
+      expect(stdout).toMatch('A fresh build may take longer than usual');
```
grabbou: Isn't it better to use `toMatchSnapshot` here as well? (just like we do for start message)

thymikee: It is, but Android command produces some kind of warning with different wording on different machines, so I'd like to keep it this way for now. I can revisit later though

thymikee:
Feedback addressed

Issue #204, opened 2017-08-01
Silent failure on build errors


joearasin:
This is a bit of a weird one -- but on some packaging errors (i.e. if I am exporting a generator, for example, which chokes React Native's build), the error gets silently swallowed, with haul outputting "Built successfully!", but opening the IOS simulator reveals "Module AppRegistry is not a registered callable module (calling runApplication)".

To reproduce, drop something along the lines of:
```javascript
export function* kaboom() {
  console.log("Boom");
} 
```
into a file that is getting packaged by haul. 

My shell ends up outputting something along the lines of: 
```
GET /index.ios.bundle 200 3936836 - 14.583 ms
GET /index.ios.bundle 304 undefined - 5.852 ms
```
That 304 does not show up when things run smoothly.

### Haul Configuration (webpack.haul.js)

```javascript
const fs = require("fs");

module.exports = ({ platform }, defaults) => ({
  entry: `./index.${platform}.js`,
  resolve: {
    ...defaults.resolve,
    modules: [
      fs.realpathSync("node_modules"),
      fs.realpathSync("../common/node_modules"),
    ],
  },
});
```

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-beta.2
| react-native     |  0.44.0
| node             |  8.1.4
| npm or yarn      |  yarn 0.27.5

Krizzu:
Hey @joearasin, thanks for filling this issue.

I've created a PR #205 to fix this problem.


All best.

joearasin:
Awesome. thanks for the quick fix. I think the changes in today's release (maybe #183) actually fixed my issue, but anything helps!

Krizzu:
@joearasin 

Glad you have your problem solved.

Cheers :)

PR #205, opened 2017-08-01
Update react-native preset

Krizzu:
This PR addresses #204. The problem was in a react-native preset, where regenerator-runtime was undefined.

Big props to @zamotany for nailing down the problem 👍

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
     "babel-plugin-transform-flow-strip-types": "^6.22.0",
     "babel-plugin-transform-object-rest-spread": "^6.23.0",
     "babel-preset-env": "^1.2.2",
-    "babel-preset-react-native": "^1.9.1",
     "babel-register": "^6.24.0",
     "camelcase-keys": "^4.0.0",
     "case-sensitive-paths-webpack-plugin": "^2.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -64,7 +63,11 @@
     "flow-bin": "^0.42.0",
     "husky": "^0.13.3",
     "jest": "^19.0.2",
-    "lint-staged": "^3.4.0"
+    "lint-staged": "^3.4.0",
+    "babel-preset-react-native": "^2.0.0"
+  },
+  "peerDependencies": {
+    "babel-preset-react-native": "^2.0.0"
```
satya164: Do we need to put this in peer dependencies instead of dependencies?

zamotany: Yes, we don't want to bump the preset version each time it changes manually, when you `react-native init` you will have it installed so why would the haul need to install it itself? Besides `preset-react-native` is tied to RN version, so we would want to use the one provided with RN, otherwise we will have constantly getting issues that sth doesn't work and it's probably because of outdated RN preset.

satya164: > we don't want to bump the preset version each time it changes manually

That's fair, but do we need a peer dependency? React Native doesn't have it as a peer dep AFAIK, so it's one extra step

PR #209, opened 2017-08-03
Unit tests for CLI entry

thymikee:
**Summary**

After #200 is merged (which is current base of this PR – to be changed), I'd like to add some more tests. Here are tests for CLI entry.

### DIFF
--- a/src/__tests__/logger.test.js
+++ b/src/__tests__/logger.test.js
@@ -1,0 +1,34 @@
+/**
+ * Copyright 2017-present, Callstack.
+ * All rights reserved.
+ *
+ * @flow
+ */
+
+import stripAnsi from 'strip-ansi';
+import logger from '../logger';
+
+beforeEach(() => {
+  // $FlowFixMe
+  console.log = jest.fn();
+});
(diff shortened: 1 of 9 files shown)


src/__tests__/logger.test.js
```diff
@@ -0,0 +1,46 @@
...
+  const { calls } = console.log.mock;
+  const logs = calls.map(call => call.join(' ')).join('\n');
+  expect(stripAnsi(logs)).toMatchSnapshot();
+});
+
+test('debug logs are displayed without extra newlines', () => {
+  logger.debug('pigeon', 'no space above');
+  logger.debug('pigeon', 'no space above, but has newline\n');
+  logger.debug('pigeon', 'space above');
+  const { calls } = console.log.mock;
```
grabbou: To be consistent, we should separate `logger` (line 42) with line 43 by adding an empty line? See line 33.

thymikee: Addressed.

thymikee:
You've merged it to `integration-tests` branch. I'll send another PR which is rebased to master

PR #211, opened 2017-08-04
Test resolvers

thymikee:
Blocked by #200


grabbou:
Let's merge other PRs first so the diff is clearer.

grabbou:
@satya164 I am leaving you that one to review since you have authored the resolver logic.

PR #224, opened 2017-08-23
Reload button

Krizzu:
Small feature that allows to reload your App(s), by using `Reload` button at debugger. 
This uses Live Reload feature, so in order to make it work on Your device, make sure Live Reload is active.

### DIFF
--- a/src/server/middleware/liveReloadMiddleware.js
+++ b/src/server/middleware/liveReloadMiddleware.js
@@ -8,6 +8,18 @@
  */
 function liveReloadMiddleware(compiler) {
   let watchers = [];
+  const headers = {
+    'Content-Type': 'application/json; charset=UTF-8',
+  };
+
+  function notifyAllWatchers() {
+    watchers.forEach(watcher => {
+      watcher.res.writeHead(205, headers);
+      watcher.res.end(JSON.stringify({ changed: true }));
+    });
+
+    watchers = [];
+  }
 
   return (req, res, next) => {
     /**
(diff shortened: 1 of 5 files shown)


Krizzu:
This is actually a WIP as I want to let user choose if he wants to reload on save, reload by pressing a button or both.

src/server/middleware/liveReloadMiddleware.js
```diff
@@ -6,21 +6,19 @@
 /**
  * Live reload middleware
  */
-function liveReloadMiddleware(compiler) {
-  let watchers = [];
```
grabbou: Can we revert this change? I would like watchers to be declared within the middleware itself. I see that primary reason for adding it was to share it between this middleware and `reloadAppMiddleware` (singleton). In this case, I believe you can just add respective if statement (`if (req.path === '/reloadapp')`) below the one declared on line 17. It's okay for one middleware to operate on more than one path (see `devToolsMiddleware.js`)

src/server/middleware/reloadAppMiddleware.js
```diff
@@ -0,0 +1,31 @@
...
+
+function reloadAppMiddleware(req, res, next) {
+  /**
+   * On reload request, notify all watchers
+   */
+  if (req.path === '/reloadapp') {
+    const headers = {
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+    Watchers.getList().forEach(watcher => {
```
grabbou: Nice, but `getList()` is redundant here. `watchers` could be just an array.

src/server/util/WatcherList.js
```diff
@@ -0,0 +1,28 @@
...
+  watcherList.splice(index, 1);
+};
+
+const getList = () => watcherList;
+
+const clearList = () => {
+  watcherList.length = 0;
+};
+
+module.exports = {
```
grabbou: You get my points by writing this abstraction, but in case of this PR, I believe previous version using an array and primitive operation was a bit better.

Krizzu:
Hey @grabbou thanks for the feedback!
I'll send changes by the end of this week

Krizzu:
Hey again,
I've changed implementation of my feature. Also added option `reload` to reload devices from command line.

grabbou:
Looks good.

PR #225, opened 2017-08-24
Fix sourcemaps in dev

satya164:
cc @zamotany

### DIFF
--- a/src/utils/makeReactNativeConfig.js
+++ b/src/utils/makeReactNativeConfig.js
@@ -66,7 +66,7 @@ const getDefaultConfig = ({
    * `cheap-module-source-map` is faster than `source-map`,
    * but it doesn't have column mappings
    */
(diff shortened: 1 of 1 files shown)


src/utils/makeReactNativeConfig.js
```diff
@@ -66,7 +66,7 @@ const getDefaultConfig = ({
    * `cheap-module-source-map` is faster than `source-map`,
    * but it doesn't have column mappings
    */
-  devtool: dev ? 'cheap-module-source-map' : 'source-map',
+  devtool: dev ? 'eval-source-map' : 'source-map',
```
zamotany: And what should the alternative be? `source-map` does not work, neither `cheap-eval-source-map`

grabbou:
I am going to merge this and review before releasing. Thank you!

PR #229, opened 2017-09-04
Bundle version served by dev-server to index.${platform}.js

### LINKED ISSUE #228: React-native 0.48.1 content-type issue
I just upgraded what is effectively a Hello World react-native app + Haul to react-native 0.48.1, and now upon launching I am getting an error in my simulator:

> Expected Content-Type to be 'application/javascript' or 'text/javascript' but got 'application/octet-stream; charset=UTF-8'

My assumption is that https://github.com/facebook/react-native/blob/v0.48.1/React/Base/RCTJavaScriptLoader.mm#L247 is tripping on something in haul, as running `curl -v -D - http://localhost:8081/index.ios.bundle -o /dev/null` indicates that the content-type of my bundle is indeed `Content-Type: application/octet-stream; charset=UTF-8`.

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-beta.5 
| react-native     |  0.48.1
| node             | 8.4.0
| npm or yarn      | 0.27.5

joearasin:
Fixes #228 pending fix of https://github.com/facebook/react-native/issues/15791 (perhaps by way of merge of https://github.com/facebook/react-native/pull/15792 )

Going to want to give it a spin when moving parts are settled / write tests, but this is a general approach that should work.

### DIFF
--- a/src/commands/start.js
+++ b/src/commands/start.js
@@ -32,6 +32,7 @@ async function start(opts: *) {
       dev: opts.dev,
       minify: opts.minify,
       port: opts.port,
+      serve: true,
     },
   );
 
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -27,6 +27,7 @@ const loggerMiddleware = require('./middleware/loggerMiddleware');
 const missingBundleMiddleware = require('./middleware/missingBundleMiddleware');
 const systraceMiddleware = require('./middleware/systraceMiddleware');
 const rawBodyMiddleware = require('./middleware/rawBodyMiddleware');
+const rewriteUrl = require('./middleware/rewriteUrl');
 
 const WebSocketServer = require('ws').Server;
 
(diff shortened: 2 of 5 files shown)


joearasin:
and by `index.${platform}.js` in the commit message I mean `index.${platform}.bundle.js`

satya164:
Instead of rewriting the URL, can't we set the content type?

joearasin:
Is there a way to set it such that it does not get smashed by webpack-dev-middleware setting it? If so, it would certainly be a preferable solution

joearasin:
Apparently webpack-dev-middleware has such a flag that I missed yesterday: https://github.com/webpack/webpack-dev-middleware/pull/216#issuecomment-327147606

joearasin:
Closing b/c #230 is a better solution.

PR #230, opened 2017-09-05
Set mimetype for bundle files

joearasin:
I failed at reading the webpack-dev-server code. It exposes a parameter for adding mime-type detection. This ensures that the `.bundle` file is served with an `application/javascript` content-type.

https://github.com/callstack-io/haul/pull/229 is unnecessary, but things are still broken pending fix of https://github.com/facebook/react-native/issues/15791

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -51,6 +51,7 @@ function createServer(
     reporter: null,
     stats: 'errors-only',
     hot: true,
+    mimeTypes: { 'application/javascript': ['bundle'] },
     watchOptions: {
       aggregateTimeout: 300,
       poll: 1000,


joearasin:
Addresses #228

appden:
Unfortunately the `mimeTypes` option [updates the global mime lookup module](https://github.com/webpack/webpack-dev-middleware/blob/master/lib/Shared.js#L36), which isn't great _and_ has the issue you mentioned where `charset=UTF-8` is included which confuses React Native.

Instead you could pass an override in the [`headers` option](https://github.com/webpack/webpack-dev-middleware/blob/master/middleware.js#L74-L78), which would take precedence over the default generated `Content-Type` header.

appden:
Never mind, that would screw with assets served from the middleware. Ignore me. :smile:

PR #237, opened 2017-09-11
Multiple progress bars

### LINKED ISSUE #208: In mode `all`, progress indicator is broken
Progress indicator is receiving two, mutually exclusive, progress events, one from iOS compiler and second, from Android compiler.

We should build a nice proxy / abstraction that will collect all events from compilers and report a summary.

Krizzu:
Fixes #208.

screen shot 2017-09-11 at 15 14 54
(this white rectangle is my terminal cursor).


grabbou:
Oh, that's dope! One tiny suggestion: Can we just show a progress bar w/o `ios` or `android` if we are running in a single platform mode?

screen shot 2017-09-14 at 13 14 03

src/utils/haulProgressBar.js
```diff
@@ -0,0 +1,53 @@
...
+const barOptions = {
+  complete: chalk.bold('='),
+  incomplete: ' ',
+  width: 20,
+  total: 100,
+};
+
+function createBarFormat(plat: string) {
+  const platform =
+    plat === 'android'
```
grabbou: This is potential issue in the future when we add more platforms. Is there any chance we can just capitalise `plat` passed?

grabbou: (let's see how it's being done in different places)

Krizzu:
@grabbou 
I made it more generic to handle future platforms.
Also, I'd suggest to keep platform label is there is only one build happening, as it can be a sanity-check in case You selected platform You did not want to (so You get feedback immediately).

grabbou:
Thanks!

> Also, I'd suggest to keep platform label is there is only one build happening, as it can be a sanity-check in case You selected platform You did not want to (so You get feedback immediately).

Well, we already provide quite explicit interface informing about the platform (and eventually want to default to `all` in the future)

screen shot 2017-09-16 at 15 10 35

Krizzu:
Fair enough. Going to do those changes next week, as few workarounds will be necessary.

satya164:
Let's get this merged soon to avoid more merge conflicts :)

src/utils/makeReactNativeConfig.js
```diff
@@ -58,168 +57,170 @@ const getDefaultConfig = ({
...
-        ? [
-            new webpack.optimize.UglifyJsPlugin({
+    },
+    plugins: [
+      /**
+       * MacOS has a case insensitive filesystem
+       * This is needed so we can error on incorrect case
+       */
+      new CaseSensitivePathsPlugin(),
+      // new ProgressBarPlugin({
```
satya164: remove commented out code

Krizzu:
Let's merge this, and create an issue to hide `platform` name if in single-build mode

src/utils/makeReactNativeConfig.js
```diff
@@ -58,168 +57,170 @@ const getDefaultConfig = ({
...
+        process.env.NODE_ENV === 'production'
+          ? []
+          : new webpack.HotModuleReplacementPlugin(),
+      )
+      .concat(
+        minify
+          ? [
+              new webpack.optimize.UglifyJsPlugin({
+                /**
              * By default, uglify only minifies *.js files
```
satya164: comment indentation sems wrong

Issue #207, opened 2017-08-03
Better error when something is already listening on port


grabbou:
When there's e.g. another `packager` listening at another port, instead of printing an error:

```
INFO  Running haul start --platform all
events.js:163
      throw er; // Unhandled 'error' event
      ^

Error: listen EADDRINUSE :::8081
    at Object.exports._errnoException (util.js:1034:11)
    at exports._exceptionWithHostPort (util.js:1057:20)
    at Server._listen2 (net.js:1265:14)
    at listen (net.js:1301:10)
    at Server.listen (net.js:1392:7)
    at Object.start [as action] (/Users/grabbou/Repositories/haul/src/commands/start.js:92:7)
    at run (/Users/grabbou/Repositories/haul/src/cliEntry.js:177:19)
```

either:
a) print a better / nicer error message
b) ask user what to do (exit / kill process and restart)

I am in favour of `b` which aligns well with our general idea of fixing mistakes.

thymikee:
Why not try to automatically connect to another free port, like Browsersync does?

satya164:
@thymikee the client would need to know to connect to the new port

grabbou:
Yeah, if we change the port transparently, one will have to re run it's native client which I assume is not wanted behaviour

thymikee:
Fair enough, thanks!

Krizzu:
I'd like to work on this one, if no one took it already :)

PR #238, opened 2017-09-12
Handle situation when server port is in use

Krizzu:
This feature gives user a choice what to do when server port (default) 8081 is already in use. We can kill the process that uses the port and proceed with startup or simply quit.
Fix for #207 

- [X] Tested on macOS
- [x] Tested on Linux
- [x] Windows implementation
- [x] Tested on Windows

### DIFF
--- a/src/cliEntry.js
+++ b/src/cliEntry.js
@@ -12,7 +12,7 @@ const clear = require('clear');
 const inquirer = require('inquirer');
 const path = require('path');
 const chalk = require('chalk');
-const { isPortTaken, killHaulProcess } = require('./utils/haulPortHandler');
+const { isPortTaken, killProcess } = require('./utils/haulPortHandler');
 
 const pjson = require('../package.json');
 const logger = require('./logger');
@@ -184,7 +184,7 @@ async function run(args: Array<string>) {
         ],
       });
       if (userChoice === 'Quit') return;
-      const result = await killHaulProcess(options.port);
+      const result = await killProcess(options.port);
       if (!result) {
         logger.error(`Could not kill process! Aborting.`);
         return;
--- a/src/cliEntry.js
+++ b/src/cliEntry.js
@@ -12,7 +12,6 @@ const clear = require('clear');
 const inquirer = require('inquirer');
 const path = require('path');
 const chalk = require('chalk');
-const { isPortTaken, killProcess } = require('./utils/haulPortHandler');
 
 const pjson = require('../package.json');
 const logger = require('./logger');
@@ -170,28 +169,6 @@ async function run(args: Array<string>) {
   const { options, promptedOptions } = await validateOptions(opts, flags);
   const userDefinedOptions = { ...flags, ...promptedOptions };
   const displayName = getDisplayName(command.name, userDefinedOptions);
-
-  if (command.name === 'start') {
-    const isTaken = await isPortTaken(options.port);
-    if (isTaken) {
-      const { userChoice } = await inquirer.prompt({
-        type: 'list',
-        name: 'userChoice',
-        message: `Port ${options.port} is already in use. What should we do?`,
-        choices: [
-          `Kill process using port ${options.port} and start server`,
-          'Quit',
-        ],
-      });
-      if (userChoice === 'Quit') return;
-      const result = await killProcess(options.port);
-      if (!result) {
-        logger.error(`Could not kill process! Aborting.`);
-        return;
-      }
-      logger.info(`Successfully killed processes.`);
-    }
-  }
 
   if (Object.keys(promptedOptions).length) {
     logger.info(`Running ${chalk.cyan(displayName)}`);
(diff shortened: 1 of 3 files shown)


jukben:
I would consider usage e.g https://www.npmjs.com/package/fkill for the cross-platform kill. What do you think?

Krizzu:
@jukben No need for that, `process.kill` is all we need :)

Krizzu:
This one is good to go.

src/cliEntry.js
```diff
@@ -169,6 +170,28 @@ async function run(args: Array<string>) {
   const userDefinedOptions = { ...flags, ...promptedOptions };
   const displayName = getDisplayName(command.name, userDefinedOptions);
 
+  if (command.name === 'start') {
```
grabbou: I believe this should go to `start` command itself, not the `cliEntry`.

src/utils/haulPortHandler.js
```diff
@@ -0,0 +1,87 @@
...
+  return new Promise(resolve => {
+    const portTester = net
+      .createServer()
+      .once('error', () => {
+        return resolve(true);
+      })
+      .once('listening', () => {
+        portTester
+          .once('close', () => {
+            resolve(false);
```
grabbou: Can we resolve w/o waiting for `close`?
```js
srv.close();
resolve(false);
```

src/utils/haulPortHandler.js
```diff
@@ -0,0 +1,87 @@
...
+const exec = require('child_process').exec;
+
+/*
+ * Check if the port is already in use
+ */
+function isPortTaken(port: number) {
+  return new Promise(resolve => {
+    const portTester = net
+      .createServer()
+      .once('error', () => {
```
grabbou: Should we check if `err.code === 'EADDRINUSE'`

src/cliEntry.js
```diff
@@ -169,6 +170,28 @@ async function run(args: Array<string>) {
...
+      const { userChoice } = await inquirer.prompt({
+        type: 'list',
+        name: 'userChoice',
+        message: `Port ${options.port} is already in use. What should we do?`,
+        choices: [
+          `Kill process using port ${options.port} and start server`,
+          'Quit',
+        ],
+      });
+      if (userChoice === 'Quit') return;
```
grabbou: Let's avoid inline `if` statements, prefer multi-line.

src/utils/haulPortHandler.js
```diff
@@ -0,0 +1,87 @@
...
+
+      const PID = PIDInfo[index];
+
+      /*
+       * Kill process
+       */
+      try {
+        process.kill(PID);
+      } catch (e) {
+        resolve(false);
```
grabbou: You should resolve with an `error`, instead of `false`. That way, in `await` you can wrap it with try/catch instead of using `if/else` result.

grabbou: *reject with an error

src/cliEntry.js
```diff
@@ -169,6 +170,28 @@ async function run(args: Array<string>) {
...
+      const result = await killProcess(options.port);
```
grabbou:
```js
try {
  await killProcess();
} catch () {
}
```

grabbou:
Oh, I've actually left a couple of changes more

src/utils/haulPortHandler.js
Krizzu: First error we'll get when starting a server is `EADDRINUSE`, so I think we can safely skip this

Krizzu:
@grabbou 
Okay, moved logic to `start` command and changes few bits, according to feedback.

grabbou:
Looks good to me. @satya164 please let me know what you think and if you are happy, merge it.

src/utils/haulPortHandler.js
```diff
@@ -0,0 +1,84 @@
...
+
+      const PID = PIDInfo[index];
+
+      /*
+       * Kill process
+       */
+      try {
+        process.kill(PID);
+      } catch (e) {
+        reject(e.toString());
```
satya164: Why reject with a string?

src/commands/start.js
```diff
@@ -7,19 +7,41 @@
...
+      choices: [`Kill process using port ${opts.port} and start Haul`, 'Quit'],
+    });
+    if (userChoice === 'Quit') {
+      process.exit();
+    }
+    try {
+      await killProcess(opts.port);
+    } catch (e) {
+      logger.error(`Could not kill process! Reason: \n ${e}`);
+      process.exit();
```
satya164: we probably want `process.exit(1)` here

src/utils/haulPortHandler.js
Krizzu: We don't need stack trace here, so simple error message should be enough

satya164: @Krizzu yea, but we're handling it when we catch the error, right? also if only message is the intention, `e.message` would be better, coz `e.toString()` will prepend `Error:`

PR #239, opened 2017-09-22
Support multiple files in webpack config

tqc:
Adds support for webpack configs that generate multiple files in the form

    entry: {
        "package1": defaults.entry.concat("./package1.js"),
        "package2": defaults.entry.concat("./package2.js")
    },
    output: {
        path: path.join(__dirname, "dist"),
        filename: "[name].bundle"
    },

Also a couple of minor updates from debugging - lint issues that were blocking the tests and improved logging on bundle to catch minification  errors.

### DIFF
--- a/src/commands/bundle.js
+++ b/src/commands/bundle.js
@@ -64,8 +64,21 @@ async function bundle(opts: *) {
   const stats = await new Promise((resolve, reject) =>
     compiler.run((err, info) => {
       if (err || info.hasErrors()) {
+        if (err) {
+          logger.error(err);
+        }
+        if (info && info.hasWarnings()) {
+          info.toJson().warnings.map(warning => logger.warn(warning));
+        }
+        if (info && info.hasErrors()) {
+          info.toJson().errors.map(error => logger.error(error));
+        }
+
         reject(new MessageError(messages.bundleFailed()));
       } else {
+        if (info && info.hasWarnings()) {
+          info.toJson().warnings.map(warning => logger.warn(warning));
+        }
         resolve(info);
       }
     }),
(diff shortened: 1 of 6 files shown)


tqc:
Looks like the lint errors are related to a change in prettier - The clean install I have locally uses 1.7, the build server uses 1.5, and both match ^1.5.3 from package.json.

grabbou:
Hey, would you mind providing a bit context why you need this change? From my perspective, there are other ways to implement this with already implemented features. Before providing explanations though, I would love to hear a bit more about your motivation.

tqc:
My project has multiple targets - same native code, different js entry point.

Based on the webpack docs, I set up entry as an object with multiple keys, and ran into the assumption that it is a string.

grabbou:
Thanks! You are right, we have missed that.

I've just discussed this with the team and we've figured out that it's best to remove our default `entry`, so that you can pass whatever you want (since we don't have a default, there's nothing to merge). Instead, use a plugin to inject the polyfill. 

Does that sound like an interesting approach? It's a bit different and unfortunately, that means we would have to close your PR. But, it served as a great inspiration for us :)

tqc:
Wouldn’t that be a breaking change? I was going for leaving the functionality unchanged for strings, which is probably still the most common usage.

grabbou:
Not at all - you still can pass either a string, array or a map of two previous - it's just the way Webpack works :)

tqc:
So you mean the current merge behavior is a noop anyway? That would make sense then.

tqc:
Ah, you mean a new plugin that applies a change to whatever output webpack is set to generate. Yes, agree that that is a cleaner approach than changing the config before the webpack call.

grabbou:
CC: @Krizzu how is progress on this? I believe we should land a PR that removes our custom entry manipulation in favour of a plugin.

Krizzu:
@grabbou I'm on it, will try to make PR till the end of the week

PR #241, opened 2017-10-03
Allow to find and merge .babelrc into default config

### LINKED ISSUE #231: Can't use modified .babelrc
This is part of lerna setup so I'm assuming some of the issues might have to do with that. It seems if we add a .`babelrc` (even the default with the react-native preset) EmitterSubmission fails to resolve. Any clues?

zamotany:
Fixes #231


satya164:
btw we probably can remove line `21`?

PR #244, opened 2017-10-05
Set CORS header for remote debugging

SEAPUNK:
This will allow the remote debugger to make requests to the dev server when browser's remote debugger address and dev server address differ.

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -52,7 +52,10 @@ function createServer(
     stats: 'errors-only',
     hot: true,
     mimeTypes: { 'application/javascript': ['bundle'] },
-    headers: { 'Content-Type': 'application/javascript' },
+    headers: {
+      'Content-Type': 'application/javascript',
+      'Access-Control-Allow-Origin': '*',
+    },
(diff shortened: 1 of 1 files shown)


Krizzu:
Hey @SEAPUNK, thanks for this PR!
One question: wouldn't it work if we add `headers` property [here](https://github.com/callstack/haul/blob/master/src/server/index.js#L52) with said headers as value? Do you mind giving it a go?

You can find example at [webpack-dev-middleware readme](https://github.com/webpack/webpack-dev-middleware)

zamotany:
Btw there is already existing [CORS middleware](https://github.com/expressjs/cors).

satya164:
Can you update the pr with the cors middleware?

SEAPUNK:
Yes, hold on.

SEAPUNK:
@zamotany I'll take the `headers` property approach; I find adding yet another dependency just for a single header is completely unnecessary.

SEAPUNK:
@Krizzu done

PR #246, opened 2017-10-07
Fix 0.49+ breaking changes

Krizzu:
Added a module to determine version of React Native. Based on that, serve bundles at correct path


src/utils/makeReactNativeConfig.js
```diff
@@ -65,7 +69,7 @@ const getDefaultConfig = ({
   devtool: bundle ? 'source-map' : 'eval-source-map',
   output: {
     path: path.join(root, 'dist'),
-    filename: `index.${platform}.bundle`,
+    filename: rnVersion >= 49 ? `index.bundle` : `index.${platform}.bundle`,
```
satya164: Instead of this, can you use the `semver` module to compare the versions?

src/utils/getReactNativeVersion.js
```diff
@@ -0,0 +1,32 @@
...
+const fs = require('fs');
+const logger = require('../logger');
+
+/**
+* Search top-most node_modules for React Native
+* and return it's version
+*/
+
+module.exports = (): string => {
+  const pckJsonLocation = path.join(
```
satya164: `require.resolve('react-native/package.json')` is probably better

Krizzu: I'd insist on using `cwd` to get it

satya164: I'd suggest using https://github.com/sindresorhus/resolve-cwd

Krizzu: If you insist :)

integration_tests/__tests__/__snapshots__/start.test.js.snap
```diff
@@ -1,8 +1,3 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`start command displays "Select platform" message 1`] = `
-"? Select platform to bundle for (Use arrow keys)
-❯ ios - Serves iOS bundle 
-  android - Serves Android bundle 
-  all - Serves both platforms"
-`;
+exports[`start command displays "Select platform" message 1`] = `""`;
```
Krizzu: Something's not right here. Investigating now

PR #242, opened 2017-10-03
Upgrade to webpack 3

zamotany:
Closes #201

PR #213, opened 2017-08-08
Add tests for some utils

thymikee:

PR #252, opened 2017-10-13
 chore: bump flow to 0.57.2

thymikee:
**Summary**

We had quite an outdated Flow, so this PR includes the latest.
Also because github failed to merge snapshots after #213 and #242, I had to re-record them, so this fixes CI btw.

### DIFF
--- a/src/hot/client/hotApi.js
+++ b/src/hot/client/hotApi.js
@@ -1,7 +1,7 @@
 /**
  * Copyright 2017-present, Callstack.
  * All rights reserved.
- * 
+ *
  * @flow
  */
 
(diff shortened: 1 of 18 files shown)


src/hot/client/hotApi.js
```diff
@@ -17,24 +17,25 @@ import resetRedBox from './utils';
...
  * Wrap root component factory with custom HotWrapper, which allows for deep force update
  * and servers as an error boundry.
  */
 export function makeHot(initialRootFactory: Function, id?: string = 'default') {
   return () => {
-    class HotWrapper extends Component {
-      state: {
-        error: ?Object,
+    class HotWrapper extends Component<*, State> {
+      state = {
```
thymikee: Hope we support class properties?

src/loaders/assetLoader.js
```diff
@@ -28,7 +28,12 @@ module.exports = async function assetLoader() {
 
   const query = utils.getOptions(this) || {};
   const options = this.options[query.config] || {};
-  const config: Config = Object.assign({}, options, query);
+  const config: Config = Object.assign(
+    {},
+    { platform: 'ios', root: process.cwd() },
```
thymikee: `options` and `query` could be empty objects, so I had to introduce default `platform` and `root`. cc @satya164

PR #256, opened 2017-10-19
Use quotation marks for HTML param values

mitcom:
This PR use quotation mark `"` to wrap HTML param value against apostrophe `'` in `src/server/assets/debugger.html`


Krizzu:
cc: @satya164

grabbou:
Good call!

Issue #115, opened 2017-04-10
Investigate how to reduce bundle size


satya164:
The bundle generated by Haul is bigger than the one generated by packager. We need to see why it's the case. Might be just because webpack generates more code than RN packager.

PR #258, opened 2017-10-19
Use module concatenation plugin and set NODE_ENV to production

zamotany:
#115 

Bundle size went down from 1MB+ to 621KB (65KB behind metro-bundler) for fresh RN project.

We should see more size decrease if we stop transpiling ES imports/exports to CJS.


thymikee:
Update the snapshots and we're good

PR #260, opened 2017-10-23
Handle different webpack entries

Krizzu:
Now we handle different webpack entries, by injecting necessary polyfills to each entry.

Also applied changes to displaying proper entry files messages

### DIFF
--- a/src/utils/__tests__/getEntryFiles.test.js
+++ b/src/utils/__tests__/getEntryFiles.test.js
@@ -1,0 +1,57 @@
+/**
+ * Copyright 2017-present, Callstack.
+ * All rights reserved.
+ *
+ * @flow
+ */
+
+const getEntryFiles = require('../getEntryFiles');
+const path = require('path');
+const dedent = require('dedent');
+
+const fakePolyfill = 'polyfill/to/remove';
+const fakeCwd = '/usr/fake';
+
+const originalCWD = process.cwd;
+
+beforeEach(() => {
+  // $FlowFixMe
+  process.cwd = () => fakeCwd;
+});
+
+beforeAll(() => {
+  // $FlowFixMe
+  process.cwd = originalCWD;
+});
+
+describe('getEntryFiles handles', () => {
+  test('entry as a string', () => {
+    const entry = '/path/to/entry.js';
+    const result = getEntryFiles(entry);
+    expect(result).toBe(path.join(fakeCwd, entry));
+  });
+  test('entry as an array', () => {
+    // polyfill is being removed from entries
+    const entry = [fakePolyfill, 'entry/one.js', 'entry/two.js'];
+    const result = getEntryFiles(entry);
+
+    expect(result).toBe(
+      `${path.join(fakeCwd, entry[1])}\n${path.join(fakeCwd, entry[2])}`
+    );
+  });
+  test('entry as object', () => {
+    const entry = {
+      main: [fakePolyfill, 'index.js'],
+      vendor: [fakePolyfill, 'vendor.js', 'stuff.js'],
+    };
+    const result = getEntryFiles(entry);
+
+    expect(result).toBe(
+      dedent(`(chunk: main)
+        ${path.join(fakeCwd, entry.main[1])}
+        (chunk: vendor)
+        ${path.join(fakeCwd, entry.vendor[1])}
+        ${path.join(fakeCwd, entry.vendor[2])}`)
+    );
+  });
+});
--- a/src/utils/__tests__/makeReactNativeConfig.test.js
+++ b/src/utils/__tests__/makeReactNativeConfig.test.js
@@ -7,7 +7,10 @@
 
 import path from 'path';
 import { replacePathsInObject } from 'jest/helpers'; // eslint-disable-line import/no-unresolved
-import makeReactNativeConfig from '../makeReactNativeConfig';
+import {
+  makeReactNativeConfig,
+  injectPolyfillIntoEntry,
+} from '../makeReactNativeConfig';
 
 test('creates config from defaults', () => {
   const webpackConfig = require('./fixtures/webpack.config.js');
@@ -32,4 +35,44 @@ test('merges existing config', () => {
   expect(replacePathsInObject(configs)).toMatchSnapshot(
     'merges existing config'
   );
+});
+
+describe('injects polyfill into different entries', () => {
+  const fakePolyfillPath = 'path/to/polyfill.js';
+
+  test('entry is a string', () => {
+    const userEntry = './src/index.js';
+    const generatedEntry = injectPolyfillIntoEntry(userEntry, fakePolyfillPath);
+
+    expect(generatedEntry.length).toBe(2);
+    expect(generatedEntry[0]).toBe(fakePolyfillPath);
+  });
+
+  test('entry is an array', () => {
+    const userEntry = ['./src/index.js', './src/module.js'];
+    const generatedEntry = injectPolyfillIntoEntry(userEntry, fakePolyfillPath);
+
+    expect(generatedEntry[0]).toBe(fakePolyfillPath);
+    expect(generatedEntry.length).toBe(3);
+  });
+
+  test('entry is an object', () => {
+    const userEntry = {
+      entry1: './src/index.js',
+      entry2: ['./src/module.js', './src/vendor.js'],
+    };
+    const expectedEntry1 = [fakePolyfillPath, './src/index.js'];
+
+    const expectedEntry2 = [
+      fakePolyfillPath,
+      './src/module.js',
+      './src/vendor.js',
+    ];
+    const generatedEntry = injectPolyfillIntoEntry(userEntry, fakePolyfillPath);
+
+    expect(generatedEntry).toMatchObject({
+      entry1: expectedEntry1,
+      entry2: expectedEntry2,
+    });
+  });
 });
--- a/src/utils/getEntryFiles.js
+++ b/src/utils/getEntryFiles.js
@@ -1,0 +1,52 @@
+/**
+ * Copyright 2017-present, Callstack.
+ * All rights reserved.
+ *
+ * @flow
+ */
+
+const path = require('path');
+
+const resolvePath = (relativeFilePathArray: string[]) =>
+  relativeFilePathArray
+    // Remove our polyfill
+    .slice(1)
+    .map(entry => path.resolve(process.cwd(), entry));
+
+const handleMultipleEntries = (multiEntry: *) => {
+  if (Array.isArray(multiEntry)) {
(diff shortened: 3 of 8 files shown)


src/utils/makeReactNativeConfig.js
```diff
@@ -226,14 +223,45 @@ function makeReactNativeConfig(
...
+    const chunkNames = Object.keys(userEntry);
+    return chunkNames.reduce((entryObj, name) => {
+      const chunk = userEntry[name];
+      if (typeof chunk === 'string') {
+        entryObj[name] = [polyfillPath, chunk];
+        return entryObj;
+      } else if (Array.isArray(chunk)) {
+        entryObj[name] = [].concat(polyfillPath, ...chunk);
+        return entryObj;
+      }
```
thymikee: This is duplicated logic from above, maybe we could reuse it?

src/utils/getEntryFiles.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * Copyright 2017-present, Callstack.
+ * All rights reserved.
+ *
+ * @flow
+ */
+
+const path = require('path');
+
+const getEntryFile = (entries: Array<string>) => {
```
thymikee: Can you add a test for this helper?

src/utils/__tests__/makeReactNativeConfig.test.js
```diff
@@ -33,3 +33,70 @@ test('merges existing config', () => {
     'merges existing config'
   );
 });
+
+describe('injects polyfill into entries, when', () => {
+  const wpConfig = {};
```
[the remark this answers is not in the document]
Krizzu: XD

src/utils/__tests__/makeReactNativeConfig.test.js
```diff
@@ -33,3 +33,70 @@ test('merges existing config', () => {
     'merges existing config'
   );
 });
+
+describe('injects polyfill into entries, when', () => {
```
grabbou: You sure this is named properly? Should be `describe("name of feature")` and then, `it("should inject")`, no? @thymikee

src/utils/makeReactNativeConfig.js
```diff
@@ -212,7 +208,8 @@ const getDefaultConfig = ({
  */
 function makeReactNativeConfig(
   userWebpackConfig: WebpackConfigFactory,
-  options: ConfigOptions
+  options: ConfigOptions,
+  polyfillPath: string = require.resolve('./polyfillEnvironment.js')
```
grabbou: This parameter is extraneous here and should be removed. This function is meant to generate RN specific Webpack config and polyfills are implementation details of this particular function. When discussing this in person, I meant this should be an argument to `injectPolyfillIntoEntry`, not to `makeReactNativeConfig`

src/utils/makeReactNativeConfig.js
```diff
@@ -226,14 +223,45 @@ function makeReactNativeConfig(
...
 
     return config;
   });
 
   return [configs, PLATFORMS];
 }
 
+/*
+ * Takes user entries for webpack.haul.js,
+ * change them to multi-point entries
```
grabbou: Isn't this too short? (You can make lines a bit longer, right?)

src/utils/makeReactNativeConfig.js
grabbou: Recursion would be nice

src/utils/__tests__/makeReactNativeConfig.test.js
thymikee: should be whatever suites best ;)

grabbou:
Yeah, in this case - I feel like `describe` doesn't play well with verbs.

src/utils/makeReactNativeConfig.js
Krizzu: In that case, I won't be able to provide fake polyfill path in tests. I could do so, be exporting `injectPolyfillIntoEntry` function and test it there

grabbou:
In makeReactConfig no, because there's no need to. You can test fake
polyfill path in the inject polyfill function, right?

Krizzu:
Okay, changed few things according to feedback (hopefully did not skip anything xP).
Also, I took care of displaying proper messages (`starting from...` type).

Tests for both added functions.

src/utils/__tests__/getEntryFiles.test.js
```diff
@@ -0,0 +1,47 @@
...
+
+const getEntryFiles = require('../getEntryFiles');
+const path = require('path');
+
+const fakePolyfill = 'polyfill/to/remove';
+const fakeCwd = '/usr/fake';
+
+beforeEach(() => {
+  // $FlowFixMe
+  process.cwd = () => fakeCwd;
```
thymikee: Can you reset `cwd` in `afterEach` hook?

src/utils/__tests__/getEntryFiles.test.js
```diff
@@ -0,0 +1,47 @@
...
+    );
+  });
+  test('entry as object', () => {
+    const entry = {
+      main: [fakePolyfill, 'index.js'],
+      vendor: [fakePolyfill, 'vendor.js', 'stuff.js'],
+    };
+    const result = getEntryFiles(entry);
+
+    expect(result).toBe(`(chunk: main)
```
thymikee: dedent

src/utils/getEntryFiles.js
```diff
@@ -0,0 +1,51 @@
+/**
+ * Copyright 2017-present, Callstack.
+ * All rights reserved.
+ *
+ * @flow
+ */
+
+const path = require('path');
+
+const handleMultipleEntries = (multiEntry: *) => {
```
thymikee:
```js
(multiEntry: Array<string> | {[key: string]: string} | string)
```

src/utils/getEntryFiles.js
```diff
@@ -0,0 +1,51 @@
...
+      const entry = multiEntry[key];
+      if (typeof entry === 'string') {
+        fileArray.push(entry);
+        return fileArray;
+      }
+
+      const tempArr = entry
+        // Don't use first entry as this is our polyfill
+        .slice(1)
+        .map((file: string) => path.resolve(process.cwd(), file));
```
thymikee: Could this be extracted to a small helper fn below? This code is duplicated

src/utils/getEntryFiles.js
```diff
@@ -0,0 +1,51 @@
...
+      fileArray.push(`(chunk: ${key})`, ...tempArr);
+      return fileArray;
+    }, []);
+  }
+
+  return multiEntry;
+};
+
+function getEntryFiles(userEntry: *) {
```
thymikee: How about better flow types?

src/utils/getEntryFiles.js
```diff
@@ -0,0 +1,51 @@
...
+const handleMultipleEntries = (multiEntry: *) => {
+  if (Array.isArray(multiEntry)) {
+    return (
+      multiEntry
+        // Don't use first entry as this is our polyfill
+        .slice(1)
+        .map(entry => path.resolve(process.cwd(), entry))
+    );
+  }
+  if (typeof multiEntry === 'object') {
```
thymikee: `null` is an object too. Can we add extra check for that?

src/utils/getEntryFiles.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  return multiEntry;
+};
+
+function getEntryFiles(userEntry: *) {
+  if (typeof userEntry === 'string') {
+    return path.join(process.cwd(), userEntry);
+  }
+  if (Array.isArray(userEntry) || typeof userEntry === 'object') {
```
thymikee: `null` once more

src/utils/makeReactNativeConfig.js
```diff
@@ -226,14 +224,45 @@ function makeReactNativeConfig(
         : userWebpackConfig
     );
 
-    // For simplicity, we don't require users to extend
-    // default config.entry but do it for them.
-    config.entry = defaultWebpackConfig.entry.concat(config.entry);
+    config.entry = injectPolyfillIntoEntry(config.entry, polyfillPath);
```
thymikee: How about merging this inside `Object.assign` above?

src/utils/makeReactNativeConfig.js
```diff
@@ -226,14 +224,45 @@ function makeReactNativeConfig(
...
+  userEntry: WebpackEntry,
+  polyfillPath: string
+): WebpackEntry {
+  if (typeof userEntry === 'string') {
+    return [polyfillPath, userEntry];
+  }
+  if (Array.isArray(userEntry)) {
+    return [polyfillPath, ...userEntry];
+  }
+  if (typeof userEntry === 'object') {
```
thymikee: I'm seeing this logic in 3rd fn already. How about creating some pattern-matching helper, and have it in one place? E.g.:
```js
return selectEntry(userEntry, {
  string: [polyfillPath, userEntry],
  array: [polyfillPath, ...userEntry],
  object: () => {...}
})
```
This could be done as a followup and labelled as a `good first issue` 🙂

src/utils/makeReactNativeConfig.js
grabbou: That would be great

thymikee:
@Krizzu keep in mind rebasing after this is merged: https://github.com/callstack/haul/pull/263

src/utils/__tests__/getEntryFiles.test.js
Krizzu: how can I do that? :D

thymikee:
```js
let originalCwd = process.cwd;

beforeEach(() => process.cwd = jest.fn(() => fakeCwd));
afterEach(() => process.cwd = originalCwd);
```

src/utils/getEntryFiles.js
Krizzu: I struggled with flow when I tried to type this entry, so I ended up using asterix

src/utils/makeReactNativeConfig.js
grabbou: Yeah, let's do it later.

grabbou:
@Krizzu where we are with this PR? For some reason, I got few notifications over the weekend and was wondering if there's been any activity since then

src/utils/getEntryFiles.js
Krizzu: Same as above, I struggled with flow, so left it with inheritance

Krizzu:
Sorry for delay, had busy weekend!

grabbou:
No worries, weekend is for fun ;) Just wanted to check the status.

sdyxch:
when merge this?

Issue #275, opened 2017-11-08
Built successfully in NaNs! When building All (iOS & Android)


Titozzz:
### Current Behavior
Built successfully in NaNs!

### Expected Behavior
Built successfully in 5.023s!

### Haul Configuration (webpack.haul.js)
Build for iOS AND Android at the same time

When building for all, build time = NaN

Krizzu:
Fixed in #277

PR #277, opened 2017-11-08
Fix build time display

Krizzu:
#275

### DIFF
--- a/src/messages/bundleBuilt.js
+++ b/src/messages/bundleBuilt.js
@@ -8,7 +8,15 @@ import type { WebpackStats } from '../types';
 
 const chalk = require('chalk');
 const dedent = require('dedent');
+
 const path = require('path');
+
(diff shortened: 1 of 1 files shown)


Titozzz:
Cool 👍

src/messages/bundleBuilt.js
```diff
@@ -8,8 +8,19 @@ import type { WebpackStats } from '../types';
...
 const chalk = require('chalk');
 const dedent = require('dedent');
+
 const path = require('path');
 
+const getBuildTime = webpackStats => {
+  const stats = webpackStats.toJson({ timing: true });
+  return stats.time
+    ? stats.time
+    : stats.children.reduce(
```
jukben: What about to go for something like this: `Math.max(...stats.children.map(({time}) => time))`? I think it's more readable.

PR #279, opened 2017-11-08
Set proper environment for remote debugging

### LINKED ISSUE #278: Lottie + Haul dosn't work while remote debug is on
The problem is related to [Lottie](https://github.com/airbnb/lottie-ios)

### Current Behavior
When you start to debug remotely it fails with an error `Cannot create property 'runCommand' on string 'LottieAnimationView'`

### Expected Behavior
Debug remotely should be possible

```js
module.exports = ({ platform }) => ({
  entry: `./index.${platform}.js`,
})
```
| software         | version
| ---------------- | -------
| Haul             |  v1.0.0-beta.9
| react-native     |  0.49.3
| node             | v8.6.0
| npm or yarn      | 1.2.1
| lottie-react-native | 2.2.7

jukben:
I've recently (#278) run into troubles with Haul and [Lottie](https://github.com/airbnb/lottie-react-native).

The problem is that Lottie relies upon [react-native-safe-module](https://github.com/lelandrichardson/react-native-safe-module) and there is the issue. React-native-safe-module is trying to ["patch" the native module in order to prevent crashes](https://github.com/lelandrichardson/react-native-safe-module/blob/master/src/NativeSafeComponent.js#L109) but the author doesn't count on that [requireNativeComponent() could return a string](https://github.com/facebook/react-native/blob/master/Libraries/ReactNative/requireNativeComponent.js#L57) and that's the issue – when you try to modify `string primitive` property under strict mode in web worker environment you get the `Cannot create property...` error.

I'm proposing hotfix to get the same behavior like in Metro Bundler.

close #278

### DIFF
--- a/src/server/assets/debuggerWorker.js
+++ b/src/server/assets/debuggerWorker.js
@@ -11,6 +11,17 @@
 
 /* eslint-disable */
 
+/**
+ * Do not use "use strict" 
+ * (https://github.com/callstack/haul/issues/278)
+ * 
+ * The problem is that Lottie relies upon react-native-safe-module and there is the issue.
+ * React-native-safe-module is trying to "patch" the native module in order to prevent crashes
+ * but the author doesn't count on that requireNativeComponent() could return a string
+ * and that's the issue – when you try to modify string primitive property under strict
+ * mode in web worker environment you get the 'Cannot create property...' error.
+ */
+
 onmessage = (function() {
   let visibilityState;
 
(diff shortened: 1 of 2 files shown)


thymikee:
Looks like local prettier got into your way?

jukben:
@thymikee It seems that Prettier is poorly set it :(

```
$ eslint --fix .
✨  Done in 4.01s.
```

PR #280, opened 2017-11-08
feat(build): show details of warnings if any

### LINKED ISSUE #274: If project was build with warnings we should list them

jukben:
basically analogic behavior to errors reporting when a build fails

close #274


jukben:
@thymikee I think it's better to leave the default styles of the errors (lot of them are already styled, for example, those from `eslint-loader`).

Does anyone want to help me with the tests? XD

thymikee:
If they're already colored then I'm cool with it. 
As for the test, this looks like a pure function, so should be pretty straightforward – create a test file with corresponding name, e.g. `__tests__/bundleBuilt.test.js` and assert on output. You'll probably need to mock the Webpack `stats` obj though.
Also, this can be done as a followup, so not a blocker.

jukben:
I've added a simple test to check this (and others) behaviors of `bundleBuilt` message.

From now, time is shown also if there are some warnings https://github.com/callstack/haul/pull/280/files#diff-ec026bf5f69761c81c5a84c227b61750R34

satya164:
btw instead of mocking chalk you can also set `FORCE_COLOR=0` when running tests

satya164:
You could do `process.env.FORCE_COLOR = 0` in `setupTestFramework.js` too. But this is fine.

jukben:
That's true, you can mess around with `process.envs`, thanks that you pointed to this, I wasn't aware of it. 👍

PR #285, opened 2017-11-16
Compact with third party npm tool like cnpm

bakso:
Compact with third party npm tool like cnpm/npminstall, cause they use symbliclink to increase install speed, the real path of haul is `_haul-dog@1.0.0-beta.10@haul` is not match the regex statement.

### DIFF
--- a/src/utils/makeReactNativeConfig.js
+++ b/src/utils/makeReactNativeConfig.js
@@ -71,7 +71,7 @@ const getDefaultConfig = ({
         { parser: { requireEnsure: false } },
         {
           test: /\.js$/,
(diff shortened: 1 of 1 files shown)


thymikee:
I guess this can be safely merged? Would appreciate a comment about China npm though.

thymikee:
@bakso can you update the snapshots? Just run `yarn test -- -u`. We would like to merge this :)

zamotany:
I'm going to close this one as we decided to get rid of `exclude` and transpile everything in `node_modules` as `metro-bundler` does.

PR #289, opened 2017-11-17
feat(bundle): better minification

jukben:
I've found that on iOS my bundle was about 15 MB because of all of the source-maps. With this fix, it dropped to 2.3 MB.

### DIFF
--- a/src/utils/makeReactNativeConfig.js
+++ b/src/utils/makeReactNativeConfig.js
@@ -127,23 +127,28 @@ const getDefaultConfig = ({
         minimize: !!minify,
         debug: dev,
       }),
-      /**
-       * By default, sourcemaps are only generated with *.js files
-       * We need to use the plugin to configure *.bundle (Android, iOS - development)
-       * and *.jsbundle (iOS - production) to emit sourcemap
-       */
-      new webpack.SourceMapDevToolPlugin({
-        test: /\.(js|css|(js)?bundle)($|\?)/i,
-        filename: '[file].map',
-      }),
     ]
       .concat(
         dev
           ? [
               new webpack.HotModuleReplacementPlugin(),
+              new webpack.EvalSourceMapDevToolPlugin({
+                module: true,
+              }),
               new webpack.NamedModulesPlugin(),
             ]
-          : new webpack.optimize.ModuleConcatenationPlugin()
+          : [
+              /**
+               * By default, sourcemaps are only generated with *.js files
+               * We need to use the plugin to configure *.bundle (Android, iOS - development)
+               * and *.jsbundle (iOS - production) to emit sourcemap
+               */
+              new webpack.SourceMapDevToolPlugin({
+                test: /\.(js|css|(js)?bundle)($|\?)/i,
+                filename: '[file].map',
+              }),
+              new webpack.optimize.ModuleConcatenationPlugin(),
+            ]
       )
       .concat(
         minify
--- a/src/utils/makeReactNativeConfig.js
+++ b/src/utils/makeReactNativeConfig.js
@@ -59,7 +59,6 @@ const getDefaultConfig = ({
   const platformProgressBar = haulProgressBar(platform);
   return {
     context: root,
-    devtool: bundle ? 'source-map' : 'eval-source-map',
     entry: [],
     output: {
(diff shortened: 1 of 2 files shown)


src/utils/makeReactNativeConfig.js
```diff
@@ -59,7 +59,6 @@ const getDefaultConfig = ({
   const platformProgressBar = haulProgressBar(platform);
   return {
     context: root,
-    devtool: bundle ? 'source-map' : 'eval-source-map',
```
jukben: We don't need this because we have a plugin for handle it. 🌮

src/utils/makeReactNativeConfig.js
```diff
@@ -154,7 +153,7 @@ const getDefaultConfig = ({
                  * We need to use the plugin to configutr *.bundle to get minified
                  * Also disable IE8 support as we don't need it'
                  */
-                test: /\.(js|bundle)($|\?)/i,
+                test: /\.(js|(js)?bundle)($|\?)/i,
```
jukben: That was the bug on iOS, so the build bloated a lot. (*main.jsbundle* didn't pass this regexp)

satya164: Can you update the comment saying we also need to support `.jsbundle` for iOS? otherwise, it'll be confusing why the `js?` exists there

src/utils/makeReactNativeConfig.js
satya164: Have you verified that this is not needed?
- Chrome debugger should show sourcemaps
- When bundling, a sourcemap file should be generated

jukben: Oh, should be? I see. OK, let me check this once more.

jukben: It's OK now @satya164

src/utils/makeReactNativeConfig.js
```diff
@@ -151,10 +151,10 @@ const getDefaultConfig = ({
               new webpack.optimize.UglifyJsPlugin({
                 /**
                  * By default, uglify only minifies *.js files
-                 * We need to use the plugin to configutr *.bundle to get minified
-                 * Also disable IE8 support as we don't need it'
+                 * We need to use the plugin to configute *.bundle (Android) and *.jsbundle (iOS)
```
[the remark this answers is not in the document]
jukben: Yeah, the difference is in prod / dev mode. I've updated comment. Thanks.

src/utils/makeReactNativeConfig.js
```diff
@@ -130,10 +129,11 @@ const getDefaultConfig = ({
       }),
       /**
        * By default, sourcemaps are only generated with *.js files
-       * We need to use the plugin to configure *.bundle to emit sourcemap
+       * We need to use the plugin to configure *.bundle (Android, iOS - development)
+       * and *.jsbundle (iOS - production) to emit sourcemap
        */
       new webpack.SourceMapDevToolPlugin({
-        test: /\.(js|css|bundle)($|\?)/i,
+        test: /\.(js|css|(js)bundle)($|\?)/i,
```
zamotany: Why in this regex `(js)` is not optional? Like in
https://github.com/callstack/haul/pull/289/files#diff-332bac151e40e6cd74cea97441a4144bR158

src/utils/makeReactNativeConfig.js
zamotany: Does chrome devtools find and use those source maps? `eval-source-map` is there deliberately, otherwise we had problem with getting SM to work.

jukben: Yeah, It's OK. Anyway, I think there is no reason for use `devtool` when you use `SourceMapDevToolPlugin`

src/utils/makeReactNativeConfig.js
jukben: Good catch, thanks. It had to accidentally miss the character.

jukben:
@zamotany Hey. Now it should be OK. The problem was, that we added a plugin for source maps twice, one was due to `devtool` and the other one because of implicit adding of the plugin. But still, could you please confirm that I don't introduce any malfunction? Thanks in advance!

cc @satya164

Issue #287, opened 2017-11-16
Use workers to generate bundle


Krizzu:
Haul - Use worker to generate webpack bundle.

## How Haul works now
Once you start Haul, you're asked to specify if you're going to handle one or both platforms. In latter case, we're using Webpack's multi compiler mode.  To serve a bundle, we have to wait for both (or maybe more) compilations to finish. This equals to increased waiting time for bundle and possible block for future platform expanding.
Additionally, [we have to handle cases,](https://github.com/callstack/haul/blob/master/src/utils/makeReactNativeConfig.js#L211) [ where webpack config is an array](https://github.com/callstack/haul/blob/master/src/commands/start.js#L62-L64), [instead doing it in generic way.](https://github.com/callstack/haul/blob/master/src/commands/start.js#L91)

## What's wrong with this approach?
As mentioned, expanding such cases is hard, as you have to keep in mind few edge cases. We could make this part easier by handling webpack in more generic way.

## Proposal
I propose a solution: a custom express middleware, that'll mimic [webpack dev middleware](https://github.com/webpack/webpack-dev-middleware), but instead of running it in the same process as our server, we spawn a process with webpack, tell him where to find necessary data and serve just the output from it. 

With this approach, we can drop the need for handling multi-compiler case and allow us to fix/enhance Haul with generic mindset and serve files ON DEMAND.

## Details - WIP

Here I'll keep some tech and non-tech details, maybe for future development.

#### Pros
1. Generic mindset when building configs/other features
2. Bundling on demand - only specific request will trigger Webpack spawning and serving
3. Faster express starting
4. Never serve the wrong bundle*

#### Cons
1. We'll have to refactor major parts of Haul
2. Communication between fork and parent can only be with serializable objects

### Code

You can track the progress at #291. At the current stage, it's ready to check out (creates and sends bundle).

thymikee:
I'll just leave this package here, maybe useful, maybe not: [jest-worker](https://github.com/facebook/jest/tree/master/packages/jest-worker)

zamotany:
Will this allow to toggling compilations? Like let's say when we will have some interactive CLI, will this allow us to toggle compilation when haul is already running?

Krizzu:
There will be no compilations toggling, but compilation on demand - say, android bundle is requested. Middleware checks if it has forked webpack for android before. If no, spawn it and send bundle. Otherwise send a bundle.

thymikee:
Closed via https://github.com/callstack/haul/pull/339

Issue #10, opened 2017-03-23
Figure out multiple configs passed to compiler


grabbou:
Currently we create an array of configs, one for each platform. We used to pass that array to webpack compiler, but switched to passing a single one (current platform you are building) for performance reasons. 

When Webpack compiler receives an array of configs, it enters multi-config mode and handles them all w/o restarting. That allows us to serve all platform bundles at the same time, just like packager. 

Unfortunately, it has a couple of limitations:
- It builds all configs at once
- It recompiles all configs at once, even if one bundle didn't change
- It's sync - android gets build after ios
- Default devMiddleware waits for all builds to finish, even if you requested ios. That means you wait for 20 seconds even if iOS was ready after 5 seconds
- All bundles are built w/o waiting for a request (no lazy option)

All the above makes the development time significantly slower for a very specific use-case (having two simulators open at the same time). However, that use-case is important and we need to support it in a better way.

I'll chat with Webpack team to understand how it works currently, if we are correct and how to fix that. Generally speaking, we might need to write our own dev middleware that does these things. 

A solution would be to add a feature to bundle lazily as soon as you request platform bundle. That means we wouldn't bundle android until you load Genymotion and request it. However, as I said above, currently turning on `lazy` option has two issues:
- It is lazy for all configs - if you request `ios`, `android` also gets built
- When `lazy` is turned on, `watch` doesn't work (recompile on file change)

Current solution requires passing `--platform` flag.

This is very important to sort out now as it's going to be an important workflow detail - whether to pass platform flags or not.

zamotany:
I don't think there is a option to fix/make workaround for rebuilding iso and android bundle source files changed, since webpack treats it separatly, so ios compiler this that source code has changed for this bundle, and android compiler also. However if we rewrite dev middleware maybe we can implement solution that will lazily build bundle that is requested and watch it, but not build the other one, but if the other one is requested we stop watching the first one, builds the second and watch it or something similar.

zamotany:
Also, is something blocking us from having 2 parallel builds in separate threads?

ev1stensberg:
If I understood this correctly, you want to build recursively and async, not all at once? @grabbou

grabbou:
Yup, @ev1stensberg + first build lazy (on request for entry point).

ev1stensberg:
[...]

PR #291, opened 2017-11-20
[WIP] Use node workers to generate bundle

Krizzu:
You can read my [Proposal here](https://github.com/callstack/haul/issues/287).

To-do:
- Change behaviours of current Haul commands (start, init, bundle)
    - [ ] start
    - [ ] init
    - [ ] bundle
- Fix middlewares that were dependant on webpack `compiler` (live reload, symbolicate)
     - [ ] live reload
     - [ ] symbolicate
- [ ] (Worker) handle webpack errors correctly
- (Middleware) normalize queue system
     - [ ] instead of using `LISTENERS`, use `requestQueue`
     - [ ] add `flushing` method to `requestQueue`, to remove all tasks once bundle has been sent
- [ ] Add more todos, because already forgot it
- [ ] Rebase with latest master
- [ ] Add flow
- [ ] Add tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "image-size": "^0.5.1",
     "inquirer": "^3.0.6",
     "loader-utils": "^1.1.0",
+    "memory-fs": "^0.4.1",
     "minimist": "^1.2.0",
     "morgan": "^1.8.1",
     "multi-progress": "^2.0.0",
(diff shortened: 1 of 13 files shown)


thymikee:
@Krizzu looks like it will solve #10?

Krizzu:
@thymikee Yeah, pretty much. :D

PR #292, opened 2017-11-21
HMR hotfix

Krizzu:
HMR would thrown an error when it finds registered function that is on a prototype. With this patch, we just make sanity check if function is on `Object.prototype`, if so, it's not a component.


jukben:
Great! Maybe it would be nice if you could spare few words as a comment to explain in code what's going on.

zamotany:
Let's add some comment as @jukben said.

PR #297, opened 2017-11-23
Change "haul start" to "haul" as a default npm-script

thymikee:
This was bothering me since very beginning of using Haul. With `yarn` it's even unnecessary to add this, but I'm not sure if `npm` is as smart.

With `"haul": "haul"` instead of `"haul": "haul start"` inside npm-scripts, you can now invoke any haul command, not just start with some arbitrary args:

```sh
yarn haul # $ haul start
```

```sh
yarn haul start # $ haul start
```

```sh
yarn haul build --platform ios # $ haul build --platform ios
```

And so on. Also `start` is the default command, so 2 points for me.

### DIFF
--- a/fixtures/react-native-clean/package.json
+++ b/fixtures/react-native-clean/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "start": "node node_modules/react-native/local-cli/cli.js start",
     "test": "jest",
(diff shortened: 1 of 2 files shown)


grabbou:
Very clever, you have my respect 💯

PR #298, opened 2017-11-24
Add example-app fixture

### LINKED ISSUE #273: Create manual testing fixtures
There are 2 fixtures already set up as a part of integration tests.
We could move these up to `/fixtures`, add a couple of more complex ones, and use them as a manual testing examples.

This would ease the testing both new versions of Haul _and_ React Native.

thymikee:
This PR adds a base for fully-fledged example, which we can run to test before releasing.

Fixes #273 

cc @zamotany I couldn't get the HMR to work with latest RN (since 0.49 `index.js` is importing `App.js` which is main component) – it doesn't crash or anything, but just reloads the screen all the times.


thymikee:
This is up-to-date and ready to merge now. We can add some navigation or native libraries later :)

PR #306, opened 2017-11-28
Add a recipe for running detox tests

itaym2:
Add a small recipe which explains how to integrate haul with detox

### DIFF
--- a/docs/Recipes.md
+++ b/docs/Recipes.md
@@ -95,3 +95,22 @@ module.exports = ({ platform }, { module, resolve }) => ({
   },
 });
 ```
+
+## Detox
+[Detox](https://github.com/wix/detox) is a "grey box" e2e framework developed by wix.  
+It provides the ability to mock files during tests using [react-native-repackager](https://github.com/wix/react-native-repackager)  
+react-native-repackager is built for the standard react-native packager, so your mocks won't work with haul out-of-the-box. Luckily, it's easy to congiure haul (webpack, actually) to resolve the mocked files instead of the original ones during tests: 
+
+```javascript
+
+// webpack.haul.js
+
+resolve: {
+    ...defaults.resolve,
+    extensions: process.env.APP_ENV === 'detox_tests' // just set this env var before running haul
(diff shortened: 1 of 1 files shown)


docs/Recipes.md
```diff
@@ -95,3 +95,22 @@ module.exports = ({ platform }, { module, resolve }) => ({
...
+It provides the ability to mock files during tests using [react-native-repackager](https://github.com/wix/react-native-repackager)  
+react-native-repackager is built for the standard react-native packager, so your mocks won't work with haul out-of-the-box. Luckily, it's easy to congiure haul (webpack, actually) to resolve the mocked files instead of the original ones during tests: 
+
+```javascript
+
+// webpack.haul.js
+
+resolve: {
+    ...defaults.resolve,
+    extensions: process.env.APP_ENV === 'detox_tests' // just set this env var before running haul
```
satya164: Might wanna add an example below instead of the comment, e.g.

    Set the environment variable `APP_ENV` to 
    `detox_tests` when running Haul:

    ```sh
    APP_ENV=detox_tests yarn haul
    ```

docs/Recipes.md
```diff
@@ -95,3 +95,22 @@ module.exports = ({ platform }, { module, resolve }) => ({
   },
 });
 ```
+
+## Detox
+[Detox](https://github.com/wix/detox) is a "grey box" e2e framework developed by wix.
```
satya164: nit: can you add new lines between paragraphs?

docs/Recipes.md
```diff
@@ -95,3 +95,31 @@ module.exports = ({ platform }, { module, resolve }) => ({
   },
 });
 ```
+
+## Detox
```
thymikee: This should rather be
```
## Detox with `react-native-repackager` mocks
```
You can use Detox without this package. We can adjust this title later, when we have more Detox tips :)

itaym2: How about ` ## Mock files when running detox tests` ?

thymikee: Will do

Issue #308, opened 2017-12-01
Error on new project


iRoachie:
### Current Behavior

On creating a new project with haul the app shows error screen on launch

#### Steps to reproduce
* Create new react-native project (react-native init)
* Add haul `yarn add --dev haul`
* Start haul server `yarn run haul`
* Run the app in the iOS simulator `react-native run-ios`

### Expected Behavior

The app should run

### Haul Configuration (webpack.haul.js)

```js
module.exports = ({ platform }) => ({
  entry: `./index.js`,
});
```

### Your Environment

Mac OS High Sierra, iOS 11.1 simulator iPhone 6

| software         | version
| ---------------- | -------
| Haul             |  ^1.0.0-beta.10
| react-native     |  0.50.4
| node             | 8.7.0
| npm or yarn      | yarn 1.3.2, npm 5.4.2

Krizzu:
Hey @iRoachie 
Could You please tell me what error you get?
Have you run `Haul init` before `Haul start`?

iRoachie:
Sure. Yea I did.

![simulator screen shot - iphone 6 - 2017-12-01 at 06 21 55](https://user-images.githubusercontent.com/5962998/33478525-f431ca18-d65f-11e7-8a2d-d0a09ee92970.png)

jurajkrivda:
I can confirm this error in RN 0.49

jukben:
This is a bug related to Webpack. I'm still searching for a satisfactory solution. As a walk around you have to downgrade to Webpack 3.8.x.

This is related PR into Webpack itself: https://github.com/webpack/webpack/pull/5862

jukben:
Another walk around for this is just open dev tools to enforce "webworker" environment. 💪

thymikee:
I'll close it, but we need a proper fix, and release soon.

SEAPUNK:
you can just pin to an older version of webpack for now

PR #309, opened 2017-12-02
fix(webpack): Stick with Webpack 3.8.x

jukben:
Sice Webpack 3.9.0 they introduce some changes that self is used before this as global.

This is just a walkaround for issues related to #308 *Can't find variable: self* . I need to dig more into it.

Background: https://github.com/webpack/webpack/pull/5862

P.S: Maybe we can force all the version to avoid such a bug in future? What do you think?

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,10 @@
     "traverse": "^0.6.6"
   },
   "lint-staged": {
-    "*.js": ["eslint --fix", "git add"]
+    "*.js": [
+      "eslint --fix",
+      "git add"
+    ]
   },
   "babel": {
     "presets": [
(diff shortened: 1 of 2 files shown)


jukben:
@satya164 you are right, quick commit. 👍 Fixed.

Krizzu:
Is 3.8.1 confirmed to work without any issues?
@jukben

jukben:
I've tried it. Besides, the issue was introduced in 3.9.x so, I think so.

Krizzu:
Sweet, let's merge it then!

AvivRubys:
I think this is what's affecting my app as well. Any chance this fix will be released soon :) ?

SanyasiraoM:
@satya164 Any idea when this will merge and a version will release?

satya164:
cc @grabbou

grabbou:
I will release the new version today

PR #310, opened 2017-12-04
Update HMR docs to account for new structure (0.49.0+)

zamotany:


docs/hmr/guides/react-navigation.md
```diff
@@ -112,9 +106,9 @@ Now, the screens can be updated upon hot update by calling `redraw`.
 Place the following code snipped at the end of the file:
 
 ```diff
+i// index.js
```
satya164: typo? extra `i`

docs/hmr/Setup.md
```diff
@@ -1,35 +1,30 @@
 # Hot Module Replacement setup
+
+For projects created before version 0.49.0, use [this guide](https://github.com/callstack/haul/blob/740b6c3cfb51d3919c69e37935d69a4c96dec94e/docs/hmr/Setup.md).
+
+__The instructions below assume you're running RN 0.49.0 or newer.__
+
 > Jump to: [Automatic setup](#automatic-setup) | [Manual setup](#manual-setup)
 
 For smaller project, we provide automatic option for setting up Hot Module Replacement. If your project meets the following criteria, you can use our on-line setup for HMR:
```
satya164: nit: "smaller projects"

satya164: nit: "one-line"

docs/hmr/Setup.md
```diff
@@ -126,44 +103,39 @@ if (module.hot) {
 ```
 
 Now, you need to replace empty strings with the paths according to the following rules:
-* If you have a __single root component__, `module.hot.accept` first argument should be an array of paths to children components and the rest of the paths should be to __itself__:
+* If you have a __single root component__, it's better to extract the `module.hot.accept` logic to a parent file (the one that import the __root component__), then you can replace all of the empty strings with path to that component:
```
satya164: nit: the one that imports

PR #316, opened 2018-01-04
Fix HMR error on Android (w/o debugger attached) + use valid URL

### LINKED ISSUE #303: HMR does not work with React Native 0.49
I use React Native Navigation. When I do `import 'haul/hot/patch';` in my index.js I am getting 
`undefined is not an object (evaluating slot[i].key)` etc.

### LINKED ISSUE #317: HMR does not work on device
### Current Behavior
When using HMR, the host is hardcoded to `localhost` which does not load across network as the server is running on whatever IP from my machine.

Therefore the device does not pick up the server and I get an error.
Even with HMR disabled from the device, the JS bundle still fails to load due to the HMR configuration being provided as build time by Webpack.

### Expected Behavior
I can get HMR to work by hardcoding my server's IP at:
- https://github.com/callstack/haul/blob/master/src/utils/makeReactNativeConfig.js#L66
- https://github.com/callstack/haul/blob/master/src/utils/makeReactNativeConfig.js#L122

However this is of course not an ideal solution. Instead, what do you think of passing in a `host` option to the CLI? It could default to the current local IP of the machine, but that would be an enhancement.

```js
module.exports = () => ({
  entry: `./src/index.js`,
})
```

### Your Environment

I'm currently testing on an iPhone X with iOS 11. But this should be an issue for HMR on all devices.

| software         | version
| ---------------- | -------
| Haul             | v1.0.0-beta.11 from my fork for #315 plus the hardcoded values described above
| react-native     |  0.50.3
| node             | 8.5.0
| npm or yarn      | 1.3.2

### Other notes
I've tried to fix this with the proposed solution above (i.e. pass in a `host` option to CLI), however I don't seem to have all the correct details of this approach working. The end result of `makeReactNativeConfig.js` isn't building with the correct `host` (as well as passing the `host` elsewhere within `haul`). If I end up getting it all working, I'll send up a PR!

zamotany:
Fixes #303 
Fixes #317

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "react-native": "*"
   },
   "dependencies": {
+    "@zamotany/react-proxy": "3.0.0-alpha.4",
     "babel-core": "^6.24.0",
     "babel-loader": "^7.1.2",
     "babel-plugin-transform-flow-strip-types": "^6.22.0",
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "jest": "test",
     "lint-staged": "^3.4.0",
     "mkdirp": "^0.5.1",
+    "react-native": "^0.51.0",
     "rimraf": "^2.6.1",
     "traverse": "^0.6.6"
   },
--- a/src/utils/makeReactNativeConfig.js
+++ b/src/utils/makeReactNativeConfig.js
@@ -139,10 +139,10 @@ const getDefaultConfig = ({
             ]
           : [
               /**
-               * By default, sourcemaps are only generated with *.js files
-               * We need to use the plugin to configure *.bundle (Android, iOS - development)
-               * and *.jsbundle (iOS - production) to emit sourcemap
-               */
+          * By default, sourcemaps are only generated with *.js files
+          * We need to use the plugin to configure *.bundle (Android, iOS - development)
+          * and *.jsbundle (iOS - production) to emit sourcemap
+          */
               new webpack.SourceMapDevToolPlugin({
                 test: /\.(js|css|(js)?bundle)($|\?)/i,
                 filename: '[file].map',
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -82,6 +83,7 @@
     "jest": "test",
     "lint-staged": "^3.4.0",
     "mkdirp": "^0.5.1",
+    "react-native": "^0.51.0",
```
Krizzu: We have `react-native` as peer dep, do we need it as dep?

satya164: Can you move this to `devDependencies`?

src/utils/makeReactNativeConfig.js
```diff
@@ -178,6 +177,16 @@ const getDefaultConfig = ({
...
     resolve: {
+      alias:
+        process.env.NODE_ENV === 'production'
+          ? {}
+          : {
+              /**
+               * Latest `react-proxy` version does not contain try/catches from
+               * commit 981815dca250373619138c9f5aadf12295cf1b3f.
+               */
+              'react-proxy': '@zamotany/react-proxy',
```
satya164: Maybe better to just use the GH url as a dependency rather than adding an alias cc @grabbou

zamotany: Using GH url won't work. `react-proxy` is required by `react-hot-loader` and have it as a dependency. Installing `react-proxy#981815dca250373619138c9f5aadf12295cf1b3f ` won't fix anything since `react-hot-loader` will have it's own version installed. That's why we need to alias it.

zamotany: Also, installing from commit, might work might not depending if yarn and npm will dedupe the package since version will be the same, but one will be from npm one from GH. Also, we would need to transpile `react-proxy` by outselves.

thymikee: Since `react-proxy` and `react-hot-loader` are not actively maintained, I'd say it's just simpler to have a working fork and just replace it later, when they fix issues / publish changes in the upstream

Krizzu:
@satya164 Since this one is sorted, can we merge this, please?

grabbou:
That actually makes sense ,gotcha

PR #319, opened 2018-01-05
You have to eject as well

alper:
yarn haul init won't work without ejecting (because then the xcodeproj won't have been created yet).


Krizzu:
You don't have to eject anything. `react-native-cli` creates `ios` and `android` for you

zamotany:
You need to `eject` if you use `create-react-native-app`.

alper:
Ah, that explains! I was coming at this from creat-react-native-app.

PR #320, opened 2018-01-05
react-native not available

alper:
I don't have `react-native` on my path (and I do think I have most of the tools installed). yarn run ios seems to work (and that does call it underneath.

```
yarn run v1.3.2
$ react-native run-ios
```


alper:
As in the other PR, I'm starting out with create-react-native-app and then some of this stuff does not make a lot of sense.

Krizzu:
No worries,

Next time you can open an issue to discuss changes you'd like to introduce.

PR #327, opened 2018-01-10
Fix React Native 0.52+

Krizzu:
There is a new experimental feature by Metro, called Delta bundles. It's turned on by default, and currently we have no proper implementation of it yet. Need to further investigate.

Added appropriate `Limitation` entry, how to disable that feature.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ jsconfig.json
 
 # NodeJS
 node_modules/
+!**/__tests__/**/node_modules
 npm-debug.log
 yarn-debug.log
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,13 +1,26 @@
-*~
+# OS X
 .DS_Store
 
+# Others
+*~
 *.log
-/node_modules
-integration_tests/fixtures/**/node_modules
-integration_tests/fixtures/**/.happypack
 
+# VSCode
 .vscode/
+tsconfig.json
 jsconfig.json
+
+# NodeJS
+node_modules/
+npm-debug.log
+yarn-debug.log
+
+# Webpack
+.happypack
+
+# Android/IJ
 .idea
-dist
-/coverage
+.gradle
+
+dist/
+coverage/
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,8 @@
 
 *.log
 /node_modules
-fixtures/**/node_modules
+integration_tests/fixtures/**/node_modules
+integration_tests/fixtures/**/.happypack
 
 .vscode/
 jsconfig.json
--- a/.gitignore
+++ b/.gitignore
@@ -15,8 +15,6 @@ node_modules/
 npm-debug.log
 yarn-debug.log
 
-# Webpack
-.happypack
 
 # Android/IJ
 .idea
--- a/src/server/middleware/requestChangeMiddleware.js
+++ b/src/server/middleware/requestChangeMiddleware.js
@@ -23,10 +23,15 @@ module.exports = function requestChangeMiddleware(
       RN 0.52+ comes with experimental feature called Delta Bundles (from metro)
       So to overcome any issues with this feature, one needs to disable it DevMenu
     */
-    req.url = req.url.replace(
-      /index.(bundle|delta)/,
-      `index.${platform}.bundle`
-    );
+    if (platform === 'android' && /\.delta/.test(req.url)) {
+      res
+        .status(500)
+        .send(
+          'Currently Haul does not support Delta bundles. Please disable them in Dev Settings'
+        );
+    }
+
+    req.url = req.url.replace(/index.bundle/, `index.${platform}.bundle`);
   }
   next();
 };
--- a/src/server/middleware/requestChangeMiddleware.js
+++ b/src/server/middleware/requestChangeMiddleware.js
@@ -19,7 +19,14 @@ module.exports = function requestChangeMiddleware(
 ) {
   const { platform } = req.query;
   if (platform) {
-    req.url = req.url.replace('index.bundle', `index.${platform}.bundle`);
+    /*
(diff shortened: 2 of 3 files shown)


src/server/middleware/requestChangeMiddleware.js
```diff
@@ -19,7 +19,14 @@ module.exports = function requestChangeMiddleware(
 ) {
   const { platform } = req.query;
   if (platform) {
-    req.url = req.url.replace('index.bundle', `index.${platform}.bundle`);
+    /*
+      RN 0.52+ comes with experimental feature called Delta Bundles (from metro)
+      So to overcome any issues with this feature, one needs to disable it DevMenu
+    */
+    req.url = req.url.replace(
+      /index.(bundle|delta)/,
```
satya164: Why do we rewrite the URL if we need to disable delta bundles? When it's disabled, will the request still come as `index.delta`? If it comes only when enabled, does serving the full bundle instead of delta bundle work?

Krizzu: Left it here just in case someone would use debugger with delta enabled. I thought throwing an Error, telling to disable it, when requested `.delta` bundles instead. WDYT?

satya164: Yes, we should throw an error asking to delta bundles. I guess with a response code 500 the device will show the error text in a Redbox

src/server/middleware/requestChangeMiddleware.js
```diff
@@ -19,7 +19,19 @@ module.exports = function requestChangeMiddleware(
...
+    */
+    if (platform === 'android' && /\.delta/.test(req.url)) {
+      res
+        .status(500)
+        .send(
+          'Currently Haul does not support Delta bundles. Please disable them in Dev Settings'
+        );
+    }
+
+    req.url = req.url.replace(/index.bundle/, `index.${platform}.bundle`);
```
zamotany: Shouldn't it be `/index\.bundle/`??

thymikee:
ping @Krizzu

Krizzu:
Escaped that dot, g2g now

.gitignore
```diff
@@ -1,12 +1,26 @@
...
-/node_modules
-fixtures/**/node_modules
 
+# VSCode
 .vscode/
+tsconfig.json
 jsconfig.json
+
+# NodeJS
+node_modules/
```
thymikee: We sure fake `node_modules` in our tests: https://github.com/callstack/haul/blob/master/src/utils/__tests__/fixtures/node_modules/react-native/package.json
Maybe lets add something like `!__tests__/**/node_modules`

.gitignore
```diff
@@ -1,12 +1,26 @@
...
+npm-debug.log
+yarn-debug.log
+
+# Webpack
+.happypack
```
thymikee: We don't use happypack anymore. If you have the dir, you better remove it (and this entry too) 😄

.gitignore
Krizzu: I think that's not necessary, as `node_modules` in fixtures are untouched

Krizzu: Just checked, not being ignored

satya164: Yeah,`node_modules/` will ignore any directory matching this glob. we need to add `!__tests__/**/node_modules` if we don't wanna ignore it

Krizzu: `git check-ignore src/utils/__tests__/fixtures/node_modules/react-native/package.json`

zamotany: You're wrong @satya164 

Using `git check-ignore` we can see that the `./src/utils/__tests__/fixtures/node_modules/react-native/package.json` __is not ignored__ by `node_modules/` rule.

```bash
git check-ignore ./src/utils/__tests__/fixtures/node_modules/react-native/package.json # doesn't output anything -> not ignored

git check-ignore ./node_modules/react-native/package.json # outputs path -> ignored
```
and also

```bash
git check-ignore ./src/utils/__tests__/fixtures/node_modules/react-native/package.json --no-index
echo $? # gives exit code 1 -> None of the provided paths are ignored.

git check-ignore ./node_modules/react-native/package.json --no-index
echo $? # gives exit code 0 -> One or more of the provided paths is ignored.
```

satya164: If the file is already being tracked by git, it's not affected by entries in .gitignore

Krizzu: @satya164 Fair enough. Should we still add it to ignore, as it's already been tracking?

zamotany: @satya164 That's why I used `--no-index` option, it doesn't care if the file was tracked or not - the result is still the same.

satya164: @zamotany I dunno how you are testing, but if I remove `!**/__tests__/**/node_modules` from `.gitignore`, I get:

```sh
git check-ignore ./src/utils/__tests__/fixtures/node_modules/react-native/package.json --no-index # exit code is 0

git check-ignore ./src/utils/__tests__/fixtures/node_modules/react-native/package.json # exit code is 1
```

You can also test that the `.gitignore` file affects `./src/utils/__tests__/fixtures/node_modules` folder by creating a file under that folder and running `git status`, which shouldn't output any untracked files.

satya164: > @Krizzu Should we still add it to ignore, as it's already been tracking?

Yeah, I think we should. It'll be confusing to have something being tracked and while mtaching the ignore filters.

thymikee: @satya164 if you look at latest commit, you'll see this already done, I've shown @Krizzu what's the issue locally :)

satya164: @thymikee yeah, saw it. I was just trying to disprove this statement by @zamotany 😛

image

Krizzu: Let's just merge it :D

Krizzu:
@satya164 We can merge it now

satya164:
I'm on mobile, so can't, delegating to @thymikee 😛

PR #328, opened 2018-01-11
Do not use foreign Babel configs

### LINKED ISSUE #272: Recipe breaks on unexpected token
I'm doing the recipe as it is here to get my react-native ts project to work.
https://github.com/callstack/haul/blob/master/docs/Recipes.md

It then gives me:

```
   SyntaxError: /Users/alper/Documents/iOS-App/rn/webpack.haul.js: Unexpected token (4:4)
  2 |   entry: `./index.js`,
 3 |   module: {
> 4 |     ...module,
   |     ^
 5 |     rules: [
 6 |       {
 7 |         test: /\.tsx?$/,
```

Is it dying on the spread operator? 

| software         | version
| ---------------- | -------
| Haul             |  haul@1.0.0-beta.7
| react-native     |  react-native@0.49.5
| node             | v8.9.0
| npm or yarn      | yarn 1.3.2

Full haul.js:

```
module.exports = ({ platform }, { module, resolve }) => ({
  entry: `./index.js`,
  module: {
    ...module,
    rules: [
      {
        test: /\.tsx?$/,
        exclude: '/node_modules/',
        use: [
          {
            loader: 'babel-loader',
          },
          {
            loader: 'ts-loader'
          },
        ],
      },
      ...module.rules
    ]
  },
  resolve: {
    ...resolve,
    extensions: [
      '.ts',
      '.tsx',
      `.${platform}.ts`,
      '.native.ts',
      `.${platform}.tsx`,
      '.native.tsx',
      ...resolve.extensions],
  },
});
```

zamotany:
Fixes #272

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -1,9 +1,46 @@
 #!/usr/bin/env node
 
-require('babel-register')({
-  ignore: /node_modules(?!\/haul)/,
-  retainLines: true,
-  sourceMaps: 'inline',
-});
+/**
+ * Babel config is stored in `package.json` so that
+ * Jest can read it.
+ */
+const babelConfig = require('../package.json').babel;
+
+function resolveModule(moduleId, type) {
+  return require.resolve(`babel-${type}-${moduleId}`);
+}
+
+function resolve(presetsOrPlugins, type) {
+  return presetsOrPlugins.map(item => {
+    return typeof item === 'string'
+      ? resolveModule(item, type)
+      : [resolveModule(item[0], type), item[1]];
+  });
+}
+
+/**
+ * To prevent `babel-register` from reading foreign Babel
+ * configs, `babelrc` has to be set to `false` and Haul's
+ * Babel config inlined into `babel-register`'s options.
+ * 
+ * Also presets and plugins need to be resolved to
+ * Haul's `node_modules` directory, otherwise Babel would
+ * search from them in project's `node_modules`.
+ */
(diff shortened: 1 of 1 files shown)


simonbuchan:
1. Is the package.json `babel` key used elsewhere? It would be much simpler otherwise to include it directly in `bin/cli.js`.
2. The real fix, of course, is to not use `babel-register` and to build before publish. If you want babel support in `webpack.haul.js`, that's what [`interpret`](https://github.com/js-cli/js-interpret) is for.

zamotany:
@simonbuchan

1. Jest also needs access to Babel config:
https://github.com/callstack/haul/blob/3b6aaee1025ed4766e7f2acbb2f0741e97efb7dd/bin/cli.js#L3-L6

2. We had an internal discussion about it some time ago and settled on using `babel-register`. As far as I advocate transpiling before publishing, simply _not using `babel-register`_ is not a fix. Also, if we were to transpiling before publishing, there's far easier way to have Babel support in `webpack.haul.js` than using `interpret`: `babel-core` + `require-from-string` (we actually don't need it, cause it's just a simple abstraction over node's native `module`).

simonbuchan:
Oops, missed that! Jest can use a js config which could import a common Babel config, but meh.

I'll note the point of interpret is that the user can use a variety of common languages, for example `webpack.haul.ts`. Further, config files can themselves import other files which should be transpiled. But that's a whole different issue!

PR #332, opened 2018-01-15
Run adb command from path at first

### LINKED ISSUE #331: adb reverse command assumes adb is in /opt
### Current Behavior

When running haul, I see the message ` WARN  Error running adb reverse tcp:8081 tcp:8081: /bin/sh: /opt/android-sdk/platform-tools/adb: No such file or directory`

### Expected Behavior

It should correctly run adb, which is available in my `$PATH` at `/usr/bin/adb`

### Your Environment

I'm running Arch Linux, and the problem occurs with this package installed: https://www.archlinux.org/packages/community/x86_64/android-tools/

The problem does not occur if I also install this package (but this one doesn't put its binaries somewhere in the default $PATH): https://aur.archlinux.org/packages/android-sdk-platform-tools/

For me, I don't need this to work because I'm using `react-native run-android` to run my app, which independently runs `adb reverse`. I assume its necessary in some circumstance though.

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-beta.12
| react-native     |  0.52.0
| node             |  9.4.0
|  yarn      | 1.3.2

jukben:
We should keep it simple and run `adb` from a path. Assumes that user has environment setup correctly according to https://facebook.github.io/react-native/docs/getting-started.html

Fixes #331

### DIFF
--- a/src/commands/start.js
+++ b/src/commands/start.js
@@ -64,19 +64,19 @@ async function start(opts: *) {
 
   // Run `adb reverse` on Android
   if (opts.platform === 'android') {
-    const args = `reverse tcp:${opts.port} tcp:${opts.port}`;
+    const command = `adb reverse tcp:${opts.port} tcp:${opts.port}`;
 
     try {
-      await exec(`adb ${args}`);
+      await exec(command);
       logger.info(
         messages.commandSuccess({
-          command: `adb ${args}`,
+          command,
         })
       );
     } catch (error) {
       logger.warn(
         messages.commandFailed({
-          command: `adb ${args}`,
+          command,
           error,
         })
       );
--- a/src/commands/start.js
+++ b/src/commands/start.js
@@ -7,7 +7,6 @@
 import type { Command } from '../types';
 
 const webpack = require('webpack');
-const path = require('path');
 const clear = require('clear');
 const inquirer = require('inquirer');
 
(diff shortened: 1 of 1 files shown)


src/commands/start.js
```diff
@@ -66,21 +65,24 @@ async function start(opts: *) {
   // Run `adb reverse` on Android
   if (opts.platform === 'android') {
     const args = `reverse tcp:${opts.port} tcp:${opts.port}`;
-    const adb = process.env.ANDROID_HOME
-      ? `${process.env.ANDROID_HOME}/platform-tools/adb`
-      : 'adb';
+    /**
+     * ADB should be taken from path (the correct setup is even in React Native documentation)
```
satya164: This comment makes sense in the context of the PR, but won't make sense after merged, where else you should take the command from if not path? :p

Let's remove this command and `inline` the `adb` variable.

jukben: OK. I wanted to be explicit for next generations XD

src/commands/start.js
```diff
@@ -66,21 +65,18 @@ async function start(opts: *) {
   // Run `adb reverse` on Android
   if (opts.platform === 'android') {
     const args = `reverse tcp:${opts.port} tcp:${opts.port}`;
-    const adb = process.env.ANDROID_HOME
-      ? `${process.env.ANDROID_HOME}/platform-tools/adb`
-      : 'adb';
 
     try {
-      await exec(`${adb} ${args}`);
+      await exec(`adb ${args}`);
```
satya164: we can also simplify it to:

```
const command = `adb reverse tcp:${opts.port} tcp:${opts.port}`
```

and use the variable `command` everywhere instead of `adb ${args}`

jukben: Good point!

jukben:
@satya164 it should be fine now, have a look pls. 🙏

PR #333, opened 2018-01-17
Add usage instructions for npm users

dave-irvine:
Not everyone uses yarn so I've added instructions for those people still on npm.


eduardomoroni:
Actually I think this a good idea, I've decided to give a try for Haul. I don't like yarn, I'll try use your approach.

Thanks

PR #336, opened 2018-01-18
Add init script to add config to build.gradle

dave-irvine:


src/commands/init.js
```diff
@@ -115,6 +117,15 @@ async function init() {
   }
 
   await addToXcodeBuild(cwd);
+
+  const rnVersion = getReactNativeVersion(cwd);
+
+  if (semver.gte(rnVersion, '0.43.0')) {
+    await addToGradleBuild(cwd);
+  } else {
+    progress.info('Haul is not required in your build.gradle');
```
jukben: I would probably remove this info message. There is no need to inform the user that everything is OK and he should not worry about build.gradle. What do you think? @zamotany

zamotany: Yeah, I also think this message is unnecessary.

src/commands/init.js
```diff
@@ -202,6 +213,71 @@ const addToPackageScripts = async (cwd: string) => {
   );
 };
 
+const addToGradleBuild = async (cwd: string) => {
```
zamotany: Why arrow function instead of regular function? We don't gain any benefit from arrow function here and in the whole codebase we use arrow function only when it makes sense or it's needed.

src/commands/init.js
```diff
@@ -202,6 +213,71 @@ const addToPackageScripts = async (cwd: string) => {
...
+        name: 'entry',
+        message: 'Enter path to the android/app/build.gradle file',
+        validate: pathToFile =>
+          fs.existsSync(pathToFile) && pathToFile.includes('build.gradle')
+            ? true
+            : `${pathToFile} is not a valid build.gradle`,
+      },
+    ]);
+
+    entry = path.resolve(cwd, result.entry);
```
zamotany: No need for `cwd` here, `resolve` will automatically use `process.cwd()` if the path is not absolute, so either use `path.resolve(result.entry)` or `path.join(cwd, results.entry)`

src/utils/getReactNativeVersion.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @flow
+ */
+const fs = require('fs');
+const path = require('path');
+
+module.exports = function getReactNativeVersion(cwd: string) {
+  try {
+    const pak = JSON.parse(
+      fs.readFileSync(path.join(cwd, 'package.json')).toString()
```
zamotany: You can simplify the code by just `require`'ing the JSON file (it will call `JSON.parse` under the hood):
```js
const pak = require(path.join(cwd, 'package.json'));
```

src/utils/getReactNativeVersion.js
```diff
@@ -0,0 +1,27 @@
...
+    );
+
+    const deps = {
+      ...(pak.dependencies || {}),
+      ...(pak.devDependencies || {}),
```
zamotany: I would reorder this to
```js
...(pak.devDependencies || {})
...(pak.dependencies || {})
...(pak.peerDependencies || {})
```
`devDependencies` should have lesser priority than `dependencies`

src/commands/init.js
dave-irvine: Will remove

src/utils/getReactNativeVersion.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * Copyright 2017-present, Callstack.
+ * All rights reserved.
+ *
+ * @flow
+ */
+const fs = require('fs');
+const path = require('path');
+
+module.exports = function getReactNativeVersion(cwd: string) {
```
Krizzu: Would not it be better to go into `node_modules/react-native` and read `package.json` from there?

src/commands/init.js
dave-irvine: I just copied the existing addToXcodeBuild arrow function as an exemplar.

src/utils/getReactNativeVersion.js
dave-irvine: I just copied existing https://github.com/callstack/haul/blob/master/src/utils/isReactNativeProject.js

src/utils/getReactNativeVersion.js
dave-irvine: This came from existing https://github.com/callstack/haul/blob/master/src/utils/isReactNativeProject.js do you want me to refactor both?

src/commands/init.js
dave-irvine: Just borrowed from the existing addToXcodeBuild. Do you want me to refactor that instance as well?

src/utils/getReactNativeVersion.js
dave-irvine: Based this off https://github.com/callstack/haul/blob/master/src/utils/isReactNativeProject.js which I assumed was your preferred way of reading package.json to get details of packages.

src/utils/getReactNativeVersion.js
zamotany: That would be awesome 👍

src/commands/init.js
zamotany: It's more of a nit picking, but if you're fine with it, refactoring it would be nice

src/utils/getReactNativeVersion.js
Krizzu: `isReactNativeProject` only tells us if project is RN. We need exact version of RN used, so I'd say I prefer to go into node_modules

src/utils/getReactNativeVersion.js
zamotany: Especially since `require`s are cached

src/utils/getReactNativeVersion.js
zamotany: In that case you can leave it as it is.

src/commands/init.js
dave-irvine: Just trying to follow the existing conventions :)

src/utils/getReactNativeVersion.js
dave-irvine: Will change :)

src/utils/getReactNativeVersion.js
dave-irvine: Will fix in both

src/utils/getReactNativeVersion.js
jukben: Wouldn't be better if /src/utils/isReactNativeProject.js@master would return `false | string` to get rid of `getReactNativeVersion` completely? @zamotany @Krizzu

zamotany: I change `getReactNativeVersion` to return `string | null` and got rid of `isReactNativeProject`.

jukben: @zamotany that's even better!

Krizzu: Yeah, good approach @zamotany

dave-irvine:
@Krizzu @zamotany @jukben Ok I think I made all the changes that were discussed. Please re-review.

src/utils/getReactNativeVersion.js
```diff
@@ -0,0 +1,24 @@
...
+    // We deliberately want to require a dynamic string, so we have to disable flow.
+    // $FlowFixMe
+    const pak = require(path.join(
+      cwd,
+      'node_modules/react-native/package.json'
+    ));
+
+    return pak.version;
+  } catch (e) {
+    // Ignore
```
jukben: Why should we ignore that we didn't find React Native? I would like to show an error that project is likely without react-native. @zamotany @Krizzu ?

dave-irvine: I think that is up to the consumer of `getReactNativeVersion`, not the function itself.

jukben: Yeah, you are right, it will return null anyway. Which is good. 👍

src/commands/init.js
```diff
@@ -115,6 +116,13 @@ async function init() {
   }
 
   await addToXcodeBuild(cwd);
+
+  const rnVersion = getReactNativeVersion(cwd);
```
thymikee: in a followup it would be nice to move this above line 28 :)

PR #339, opened 2018-01-22
Node worker usage

Krizzu:
TODO:
- [x] tests are passing
- [x] `bundle` command is working
- [x] HMR is working
- [x] symbolicate is working
- [x] workers are not clogging system when something unexpected happens
- [x] redirect progress events from forks to parent process
- [x] document what the heck is going on there
- [x] fix crazy things happening with 2 compilations and express logging


Krizzu:
Pointing to master now? 💪

integration_tests/__tests__/start.test.js
```diff
@@ -18,17 +18,32 @@ const TEST_PROJECT_DIR = path.resolve(
...
+    const message = stripAnsi(data.toString()).trim();
+
+    if (message.length > 0) {
+      messageBuffer.push(message);
+    }
 
-  const { stdout } = runHaulSync(TEMP_DIR, ['start']);
-  expect(stripAnsi(stdout).trim()).toMatchSnapshot();
+    if (message.match(/ERROR/g)) {
+      done.fail(message);
```
thymikee: Can't we listen to Haul's events instead of matching stdout?

thymikee:
@zamotany locally the `start.test.js` is failing for me, as there's no text output while listening to stdout (just some whitespace). Mind looking into this?

integration_tests/__tests__/start.test.js
zamotany: We can, but we need to change the way we run haul.

thymikee: Would be great to do it as a followup. Feel free to disable this test for the time being.

src/compiler/Compiler.js
```diff
@@ -0,0 +1,122 @@
+/**
+ * Copyright 2017-present, Callstack.
+ * All rights reserved.
+ *
+ * @flow
+ */
+
+import type { Platform, Logger } from '../types';
+
+const EventEmitter = require('events');
```
satya164: We're using `import` in various other places. Can we change them to `require` or change these to `import` for consistency?

src/compiler/Compiler.js
```diff
@@ -0,0 +1,122 @@
...
+
+  terminate(error?: Error) {
+    this.logger.info('Shutting down Haul.');
+
+    error && this.logger.error(error.message);
+
+    Object.keys(this.forks).forEach(platform =>
+      this.forks[platform].terminate()
+    );
+    process.exit(error ? 1 : 0);
```
satya164: Maybe we shouldn't do `process.exit` in the compiler instance, assuming that this will eventually be the basis of the public API.

src/compiler/TaskQueue.js
```diff
@@ -0,0 +1,29 @@
...
+    this.nextId = 0;
+  }
+
+  add(item: any) {
+    const id = this.nextId++;
+    this.queue.push({ id, item });
+    return id;
+  }
+
+  get(id: number) {
```
satya164: Probably `get` isn't the best name as it also removes the item from the queue

src/compiler/createWebSocketServer.js
```diff
@@ -0,0 +1,29 @@
...
+const xpipe = require('xpipe');
+const http = require('http');
+const { Server } = require('ws');
+const fs = require('fs');
+
+/**
+ * Create WebSocket server using HTTP server over Unix socket.
+ */
+module.exports = function createWebSocketServer() {
+  const socketAddress = xpipe.eq(`/tmp/HAUL_SOCKET_.socket`);
```
satya164: Can we change it to a unique address? We should be able to run multiple instances of Haul with different ports. It's also necessary for expo integration.

src/compiler/createWebSocketServer.js
```diff
@@ -0,0 +1,29 @@
...
+
+  if (fs.existsSync(socketAddress)) {
+    fs.unlinkSync(socketAddress);
```
satya164: We should also delete the file when the compiler terminates.

src/compiler/Compiler.js
zamotany: Does importing a type with `require` works for u? And Haul is using `requires` everywhere.

thymikee: @satya164 let's do this in one PR with a codemod, no need to mess this now

satya164: @zamotany no, will need import 🤔
@thymikee okay

src/server/util/runAdbReverse.js
```diff
@@ -0,0 +1,35 @@
...
+
+module.exports = function runAdbReverse(port: number) {
+  const command = `adb reverse tcp:${port} tcp:${port}`;
+
+  exec(command, error => {
+    if (error) {
+      // Get just the error message
+      const message = error.message.split('error:')[1];
+
+      logger.warn(
```
zamotany: @thymikee @satya164 @Krizzu Do you have any idea where can we put a info about this in the UI?

integration_tests/utils.js
```diff
@@ -109,6 +110,25 @@ function replaceTestPath(string: string) {
   return string.replace(/^\W+(.*)integration_tests/gm, '<<REPLACED>>');
 }
 
+function fetchBundle(url: string): Promise<any> {
+  return new Promise((resolve, reject) => {
```
satya164: Looks like we have a dependency on `node-fetch`. Maybe worth using it rather than wrapping `http.get` in a promise?

src/babelRegister.js
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env node
```
satya164: Do we need this shebang? Do we ever execute this file directly?

src/compiler/Compiler.js
```diff
@@ -0,0 +1,115 @@
...
+    return Events;
+  }
+
+  forks: { [key: Platform]: Fork };
+  tasks: TaskQueue;
+
+  constructor(options: *) {
+    super();
+    this.forks = {};
+    this.tasks = new TaskQueue();
```
satya164: nit: could use the class properties for initialization too

src/compiler/Fork.js
```diff
@@ -0,0 +1,119 @@
...
+const createForkProcess = require('./createForkProcess');
+const createWebSocketServer = require('./createWebSocketServer');
+
+type ForkConstructorArgs = {
+  platform: Platform,
+  options: *,
+};
+
+const forks = {};
+let transportServer;
```
satya164: Is there a reason these two are not properties in the `Compiler` instance? Seems like when you create multiple compiler instances from a single process, it's going to reuse the same transport server (with the same address) and overwrite previous forks?

Will probably also be an issue if you had created and terminated a compiler instance previously since this variable is still present, so if you create a new compiler instance, it'll try to use it, but the connection has closed.

src/compiler/createWebSocketServer.js
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Create WebSocket server using HTTP server over Unix socket.
+ */
+module.exports = function createWebSocketServer() {
+  const socketAddress = getSocketAddress();
+
+  const httpServer = http.createServer();
+  const webSocketServer = new Server({ server: httpServer });
+
+  httpServer.listen(socketAddress);
```
satya164: Seems like this `httpServer` will never close. Will it be a memory leak if you're creating and terminating multiple compilers?

src/compiler/worker/runWebpackCompiler.js
```diff
@@ -0,0 +1,66 @@
...
+  [key: string]: string,
+  fs: Object,
+}) {
+  const emitter = new EventEmitter();
+
+  const { configPath, configOptions } = JSON.parse(options);
+  const config = getConfig(configPath, configOptions, platform);
+
+  let lastPercent = -1;
+  config.plugins.push(
```
satya164: Maybe clone the config instead of mutating it? If the user is sharing same config between multiple compilers, this could add multiple plugins to the same config.

src/compiler/worker/runWebpackCompiler.js
```diff
@@ -0,0 +1,66 @@
...
+
+  emitter.on('start', () => {
+    compiler.watch({}, () => {});
+    emitter.emit(Events.BUILD_START);
+  });
+
+  return emitter;
+};
+
+function resolveAsync(args) {
```
satya164: What does this do?

integration_tests/utils.js
zamotany: I'm in the middle of fixing `start` test suite, which also replaces this with `fetch`

src/babelRegister.js
zamotany: Nope

src/compiler/createWebSocketServer.js
zamotany: It can be, though having multiple compilers is an edge case.

src/compiler/worker/runWebpackCompiler.js
zamotany: No idea. cc: @Krizzu

src/compiler/worker/runWebpackCompiler.js
zamotany: Doesn't matter - `makeReactNativeConfig` always create a fresh copy of the config.

.flowconfig
```diff
@@ -7,5 +7,8 @@
 .*/node_modules/fbjs/*
 <PROJECT_ROOT>/fixtures/.*
 
+[untyped]
+.*/node_modules/react-stream-renderer
```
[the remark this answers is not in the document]
zamotany: #no_shame

src/compiler/worker/runWebpackCompiler.js
Krizzu: If this function is fired async, second argument is `done` callback.

zamotany:
@thymikee done

src/utils/makeReactNativeConfig.js
```diff
@@ -60,13 +57,12 @@ const getDefaultConfig = ({
...
 }): WebpackConfig => {
   // Getting Minor version
-  const platformProgressBar = haulProgressBar(platform);
   return {
     mode: dev ? 'development' : 'production',
     context: root,
     entry: [],
     output: {
-      path: path.join(root, 'dist'),
+      path: path.join(root), // removed 'dist' from path
```
satya164: > `// removed 'dist' from path`

not sure what's this!!!

integration_tests/__tests__/start.test.js
```diff
@@ -57,14 +34,48 @@ function testPlatform(platform, done) {
...
+      done();
+    }
+  }
+  _done.fail = done.fail;
+
+  testPlatform('ios', _done);
+  testPlatform('android', _done);
+});
+
+async function testPlatform(platform, done) {
```
satya164: what do you think about changing this to return a promise instead of accepting a callback? I think tests will become simpler that way.

```js
async function testPlatform(platform) {
  const res = await fetch(`http://localhost:8081/index.${platform}.bundle`);
  const bundle = await res.text();
  expect(bundle).toMatch('__webpack_require__');
}
```

Then tests will be like:

```js
test('starts server and bundling iOS platform', () =>
  testPlatform('ios');
);
 
test('starts server and bundling Android platform', () =>
  testPlatform('android');
);
 
test('starts server and bundling all platforms', () =>
  Promise.all([
    testPlatform('ios');
    testPlatform('android');
  ]);
);
```

thymikee: With current on-demand bundling, does `start --platform` even make sense? I think we should remove or adjust these tests

zamotany: Test __doesn't__ spawn `haul start` with `--platform` option. `testPlatform` is only to make a request for specific platform bundle.

thymikee: @zamotany obviously, looks like I missed that 🤦‍♂️

src/utils/makeReactNativeConfig.js
zamotany: Not sure either

integration_tests/__tests__/start.test.js
```diff
@@ -20,9 +20,13 @@ const TEST_PROJECT_DIR = path.resolve(
...
 let haul;
-beforeAll(done => {
+beforeAll(async done => {
   run('yarn --mutex network', TEST_PROJECT_DIR);
-  haul = runHaul(TEST_PROJECT_DIR, ['start']);
+  try {
+    haul = await runHaul(TEST_PROJECT_DIR, ['start']);
+  } catch (error) {
+    done.fail(error);
+  }
```
thymikee: you can get rid of the `try/catch` and `done` entirely, error will be caught normally

src/server/index.js
```diff
@@ -41,6 +41,7 @@ const WebSocketDebuggerProxy = require('./util/WebsocketDebuggerProxy');
  */
 function createServer(config: { configPath: string, configOptions: Object }) {
   const appHandler = express();
+  appHandler.disable('etag');
```
thymikee: Why is this needed?

zamotany: to remove server caching and 304s

thymikee: worth to left a comment :)

src/__tests__/__snapshots__/cliEntry.test.js.snap
```diff
@@ -6,12 +6,10 @@ Starts a new webpack server
 
 Options:
 
-  --port [number]                 Port to run your webpack server, 8081 by default
-  --dev [true|false]              Whether to build in development mode, true by default
-  --minify [true|false]           Whether to minify the bundle, 'true' by default when dev=false
-  --platform <ios|android|all>    Platform to bundle for
```
satya164: We still support this flag for bundling, right? Maybe this should be left here?

src/compiler/worker/initWorker.js
```diff
@@ -0,0 +1,83 @@
...
+const WebSocket = require('ws');
+const MemoryFileSystem = require('memory-fs');
+const mime = require('mime-types');
+
+const Events = global.requireWithRootDir('../events');
+const runWebpackCompiler = global.requireWithRootDir('./runWebpackCompiler');
+
+module.exports = function initWorker({
+  platform,
+  // fileOutput,
```
satya164: Commented out code

src/__tests__/__snapshots__/cliEntry.test.js.snap
zamotany: This is `start` command snapshot not `bundle`

src/compiler/Compiler.js
```diff
@@ -0,0 +1,147 @@
...
+      if (!Object.keys(this.forks).length) {
+        callback({
+          errors: null,
+          platform: null,
+          file: null,
+          mimeType: null,
+        });
+      }
+
+      // Callback will be invoked on `FILE_RECEIVED` event or on `FILE_NOT_FOUND`.
```
thymikee: is this comment still relevant here?

src/compiler/Fork.js
```diff
@@ -0,0 +1,121 @@
...
+    );
+
+    forks[platform] = this;
+  }
+
+  setSocket(socket: WebSocket) {
+    this.socket = socket;
+
+    // Flush enqueued messages.
+    this.enqueuedMessages.forEach(({ type, ...payload }) =>
```
thymikee: Could use `.filter` instead of `.forEach` and just return `false` each time, but meh

src/compiler/Compiler.js
zamotany: Yes

src/compiler/Fork.js
```diff
@@ -0,0 +1,121 @@
...
+      this.send(type, payload)
+    );
+    this.enqueuedMessages = [];
+
+    // $FlowFixMe
+    this.socket.addEventListener('message', ({ data }) => {
+      const { type, ...payload } = JSON.parse(data.toString());
```
thymikee: we could guard that parsing somehow, in case data is undefined?

zamotany: We could but if data is undefined the process should terminate.

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -109,32 +103,48 @@ function getRequestedFrames(req: $Request): ?ReactNativeStack {
...
+    const bundleName = `${name}.${platform}.${rest[rest.length - 1]}`;
 
     // grab our source map consumer
-    const consumer = createSourceMapConsumer(compiler);
+    const consumer = await createSourceMapConsumer(
+      compiler,
+      // $FlowFixMe
+      `http://localhost:${req.get('host').split(':')[1]}/${bundleName}.map`
+    );
+    // console.log('C', consumer);
```
thymikee: dead comment

src/server/middleware/symbolicateMiddleware.js
```diff
@@ -109,32 +103,48 @@ function getRequestedFrames(req: $Request): ?ReactNativeStack {
...
+    }
+
+    const [name, ...rest] = filename.split('.');
+    const bundleName = `${name}.${platform}.${rest[rest.length - 1]}`;
 
     // grab our source map consumer
-    const consumer = createSourceMapConsumer(compiler);
+    const consumer = await createSourceMapConsumer(
+      compiler,
+      // $FlowFixMe
```
thymikee: is this needed? line below is a string

Krizzu:
@zamotany 👏 !

PR #348, opened 2018-01-30
chore: mention automatic config setup with Gradle

thymikee:


Krizzu:
We can add note that `haul init` will do it for you and leave those notes there as well

docs/Configuring Your Project.md
```diff
@@ -20,7 +20,7 @@ Usually, a React Native project has a run script phase that runs `react-native-x
 
 ## Integrating with Gradle
 
-If you're on React Native version >= 0.43, add the following in `android/app/build.gradle` somewhere before the `apply from: "../../node_modules/react-native/react.gradle"` statement:
+If you're on React Native version >= 0.43, `haul init` will add the following code in `android/app/build.gradle` somewhere before the `apply from: "../../node_modules/react-native/react.gradle"` statement for you:
```
thymikee: @satya164 @Krizzu how about now?

satya164: Maybe something like:

If you're on React Native version >= 0.43, run the following to automatically configure your gradle config to use haul:

```sh
haul init
```

If the automatic setup didn't work for you, you can manually add the following code in android/app/build.gradle somewhere before the apply from: "../../node_modules/react-native/react.gradle" statement for you:

thymikee:
Should be better now

Krizzu:
Looks good 👍

PR #357, opened 2018-02-09
add core-js as an explicit dependency for npm users

### LINKED ISSUE #140: Cannot find module 'core-js/modules/es7.string.pad-start'
### Current Behavior
screen shot 2017-04-24 at 10 18 28 pm

### Expected Behavior
init haul correctly

can't generate configuration file

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  latest
| react-native     |  0.43.4
| node             | 7.5.0
| npm or yarn      | 4.5.0

Traviskn:
fixes #140


Traviskn:
`"core-js": "2.5.3"` is the version of core-js we installed in our project as a dev dependency to get things working, let me know if a different version or perhaps a different version range would be more appropriate

package.json
```diff
@@ -43,6 +43,7 @@
     "clear": "^0.0.1",
     "cliui": "^3.2.0",
     "clone": "^2.1.1",
+    "core-js": "2.5.3",
```
thymikee: I think we should be good with default semver range `^2.5.3` :)

Traviskn: I'll add the `^`!

PR #358, opened 2018-02-09
support bundle files without "index" in the name

Traviskn:
We have a project where we have more entry-points in addition to the `index.js`, but these other entry-point files don't have `index` in their name.  (If you're curious about the  use-case for this, it's because we are using [react-native-threads](https://github.com/joltup/react-native-threads).

This PR is a different strategy for finding the dev server origin without looking specifically for `index`.  I'm very open to feedback on this approach or if anyone has a better idea for going about it.

### DIFF
--- a/src/utils/polyfillEnvironment.js
+++ b/src/utils/polyfillEnvironment.js
@@ -42,9 +42,15 @@ const scriptURL = require('react-native').NativeModules.SourceCode.scriptURL; //
 // or Android device it will be `localhost:<port>` but when using real iOS device
 // it will be `<ip>.xip.io:<port>`. Thus the code below ensure we connect and download
 // manifest/hot-update from a valid origin.
-const match = scriptURL && scriptURL.match(/(^.+)\/index/);
+let devServerOrigin = null;
 
-global.DEV_SERVER_ORIGIN = match ? match[1] : null;
+if (scriptURL) {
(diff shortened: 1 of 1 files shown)


zamotany:
With this approach, `DEV_SERVER_ORIGIN` is missing a protocol.

Traviskn:
Oh you're right this is incomplete, I totally left out the protocol!

I was running into red screen errors, failing on this line of the polyfillEnvironment file, even though I hadn't turned on HMR. If there's a way to somehow opt out of this file getting run when I don't need HMR that would be a fine workaround for our project.

thymikee:
@Traviskn don't forget to rebase before you start working on that, we recently added a fix to this part of the codebase

zamotany:
You can move
https://github.com/joltup/haul/blob/2b7eb739cbc7d39efdc12a263a100413d350f5e4/src/utils/polyfillEnvironment.js#L39-L52

into `src/hot/client/index.js` but then it might break HMR on physical iOS devises.

Traviskn:
Yes I'll rebase on master

Traviskn:
While rebasing I noticed that the latest changes to the `polyfillEnvironment.js` actually do stop errors from being thrown, so my change isn't strictly necessary unless someone needs HMR in a bundle without index in the name.

PR #359, opened 2018-02-12
Update example-app to RN 0.53.0

czystyl:


thymikee:
This was upgraded with `react-native-git-upgrade`, right?

PR #360, opened 2018-02-12
Bump flow

zamotany:


src/hot/client/utils.js
```diff
@@ -14,6 +14,7 @@ import { Platform } from 'react-native';
 
 export default function resetRedBox() {
   if (Platform.OS === 'ios') {
+    // $FlowFixMe
     const RCTRedBox = require('NativeModules').RedBox;
```
thymikee: This is Haste – use `import {NativeModules} from 'react-native';`

PR #363, opened 2018-02-13
Fix scriptUrl for production build

### LINKED ISSUE #349: scriptURL is null when building for Release with Xcode
### Current Behavior

When building with the release build configuration in Xcode, I'm getting the following build error:
```
Terminating app due to uncaught exception 'RCTFatalException: Unhandled JS Exception: null is not an object (evaluating 't.match(/(^.+)\/index/)[1]')
```
The error is referring to the minified version of [this line](https://github.com/callstack/haul/blob/172c93acf759e43ee93f1825e683a420a9f8784c/src/utils/polyfillEnvironment.js#L45), so it seems that `NativeModules.SourceCode.scriptURL` being imported from react-native is `null`.

I'm getting this when using haul version 1.0.0-beta-12. With 1.0.0-beta-11, it works just fine (it doesn't include the line that is causing the error).

### Expected Behavior

The build should complete without error as it does in beta-11.

```
const eslintFormatter = require('react-dev-utils/eslintFormatter')
const paths = require('./paths')

  entry: `./index.js`,
  module: {
    ...defaults.module,
    rules: [
      ...defaults.module.rules,
      {
        test: /\.(js|jsx)$/,
        enforce: 'pre',
        use: [
          {
            options: {
              formatter: eslintFormatter,
            },
            loader: require.resolve('eslint-loader'),
          },
        ],
        include: paths.appSrc,
      },
      {
        test: /\.graphql$/,
        exclude: /node_modules/,
        loader: 'graphql-tag/loader',
      },
    ],
  },
})
```

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-beta-12
| react-native     |  0.52.0
| node             | 8.9.4
| npm or yarn      | yarn

czystyl:
## Summary

For production build regexp not matched anything and `scriptURL.match(/(^.+)\/index/)[1]` throws error because match was `null`, so i added statement for checking match and return null if not matched.

Fixes #349

### DIFF
--- a/src/utils/polyfillEnvironment.js
+++ b/src/utils/polyfillEnvironment.js
@@ -42,7 +42,9 @@ const scriptURL = require('react-native').NativeModules.SourceCode.scriptURL; //
 // or Android device it will be `localhost:<port>` but when using real iOS device
 // it will be `<ip>.xip.io:<port>`. Thus the code below ensure we connect and download
 // manifest/hot-update from a valid origin.
-global.DEV_SERVER_ORIGIN = scriptURL.match(/(^.+)\/index/)[1];
+const match = scriptURL.match(/(^.+)\/index/);
+
+global.DEV_SERVER_ORIGIN = match ? match[1] : null;
 // Webpack's `publicPath` needs to be overwritten with `DEV_SERVER_ORIGIN` otherwise,
 // it would still make requests to (usually) `localhost`.
 __webpack_require__.p = `${global.DEV_SERVER_ORIGIN}/`; // eslint-disable-line no-undef
--- a/src/utils/polyfillEnvironment.js
+++ b/src/utils/polyfillEnvironment.js
@@ -42,7 +42,7 @@ const scriptURL = require('react-native').NativeModules.SourceCode.scriptURL; //
 // or Android device it will be `localhost:<port>` but when using real iOS device
 // it will be `<ip>.xip.io:<port>`. Thus the code below ensure we connect and download
 // manifest/hot-update from a valid origin.
-const match = scriptURL.match(/(^.+)\/index/);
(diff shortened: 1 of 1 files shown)


src/utils/polyfillEnvironment.js
```diff
@@ -42,7 +42,9 @@ const scriptURL = require('react-native').NativeModules.SourceCode.scriptURL; //
 // or Android device it will be `localhost:<port>` but when using real iOS device
 // it will be `<ip>.xip.io:<port>`. Thus the code below ensure we connect and download
 // manifest/hot-update from a valid origin.
-global.DEV_SERVER_ORIGIN = scriptURL.match(/(^.+)\/index/)[1];
+const match = scriptURL.match(/(^.+)\/index/);
```
Krizzu: Let's do sanity check here

Krizzu: Let's do a sanity check here

Krizzu: Let's do sanity check here

`const match = scriptURL && scriptURL.match(/(^.+)\/index/);`

thymikee: Based on @czystyl findings `scriptURL` is always set, it's just different for development and production envs

Krizzu: Would it be a problem to do that sanity check tho?

Krizzu:
Don't forget to add yourself to contribution list!

src/utils/polyfillEnvironment.js
czystyl: That's true

czystyl: I added this check if RN change something.

PR #365, opened 2018-02-13
Update doc and remove old import

czystyl:
## Summary

Example doc updated

### DIFF
--- a/fixtures/example-app/index.js
+++ b/fixtures/example-app/index.js
@@ -1,4 +1,3 @@
-import 'haul/hot';
 import { AppRegistry } from 'react-native';
 import App from './App';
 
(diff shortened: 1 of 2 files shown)


fixtures/example-app/index.js
```diff
@@ -1,4 +1,3 @@
-import 'haul/hot';
```
thymikee: is it still possible to use HMR without importing this?

fixtures/example-app/index.js
```diff
@@ -1,4 +1,4 @@
...
+import 'haul/hot';
```
thymikee: remove trailing whitespace

Issue #323, opened 2018-01-10
Implement new config


zamotany:
Historically, the configuration has been done through the file `webpack.haul.js`. This proposal aims to achieve 2 things,

1. Simplify configuring commonly needed features, such as including folders for transpilation
2. Separate webpack configuration so that other tools can use it, e.g. - eslint-resolver-webpack
3. Have flexibility to configure Haul's options independent from Webpack

With the proposed approach, configuration will be done through a file called `haul.config.js`, which will look kinda like this:

```js
export default {
  // haul specific options
  webpack: {
    // webpack configuration
  }
}
```

This allows us to have the flexibility to add additional options which are unrelated to webpack, and also allows us to switch the bundler to a different one in future.

The `webpack` field can be a plain object with webpack configuration, or a function which receives various arguments:

```js
export default {
  // haul specific options
  webpack: ({ platform, production }) => ({
    // return custom config
  });
}
```

This mirrors a vanilla webpack config and it's `env` argument. This makes it so that other tools can understand the webpack config. One could also extract the webpack configuration to a separate file (e.g. - `webpack.config.js`) which other tools can read.

Currently we pass a second argument `defaults`, which is convenient, but makes the config incompatible with tools which require vanilla webpack config. We'll continue to support it, but we can also export a helper utility in case you need to support those tools:

```js
import { createWebpackConfig } from 'haul';

export default {
  webpack: createWebpackConfig(({ platform }) => {
    entry: `./index.${platform}.js`,
    include: /node_modules/,
  }),
}
```

The `createWebpackConfig` utility can take some common options has return a webpack configuration function which takes the `env` argument. To override configurations from a confguration generated this way, you could do something like this:

```js
import { createWebpackConfig } from 'haul';

export default {
  webpack: env => {
    const config = createWebpackConfig({
      entry: `./index.js`,
      include: /node_modules/,
    })(env);

    config.plugins.push(new CaseSensitivePathsPlugin());

    return config;
  },
}
```

hedgepigdaniel:
This looks useful - e.g. one useful option outside the webpack config might be to customize the webpack stats output.

simonbuchan:
So to be clear, one could do:

```js
import webpackConfig from "./webpack.config";
export default {
  webpack: env => webpackConfig({ haul: env }),
};
```

to have a single webpack config file? And this would not munge the final output, such as injecting polyfills (e.g. if I'm already injecting `core-js`), babel 6 with `react-native` preset etc?

[...]

PR #372, opened 2018-02-21
[WIP] 

jukben:
#323


thymikee:
Can you change the base to workers PR? It would be easier to review this way

jukben:
@thymikee good point I've forgotten. 👍

package.json
```diff
@@ -54,6 +55,7 @@
     "image-size": "^0.5.1",
     "inquirer": "^3.0.6",
     "loader-utils": "^1.1.0",
+    "lodash": "^4.17.5",
```
zamotany: If you really need to use `merge` from `lodash`, use `lodash.merge` module - https://www.npmjs.com/package/lodash.merge

src/utils/makeReactNativeConfig.js
```diff
@@ -224,13 +230,26 @@ function makeReactNativeConfig(
...
-      : userWebpackConfig;
+  /**
+  * Load config from file. Support both - "module.exports" and "export default {}"
+  */
+  const { webpack: webpackConfigFactory /* ...haulConfig */ } =
+    userWebpackConfig.default || userWebpackConfig;
+
+  if (
+    typeof webpackConfigFactory !== 'function' ||
+    typeof webpackConfigFactory !== 'string'
```
zamotany: I think this is better: `typeof webpackConfigFactory !== 'function' &&  typeof webpackConfigFactory !== 'object'`

src/utils/makeReactNativeConfig.js
```diff
@@ -224,13 +230,26 @@ function makeReactNativeConfig(
...
+  ) {
+    throw new Error('You have provided a wrong Webpack configuration.');
+  }
+
+  const webpackConfig =
+    typeof webpackConfigFactory === 'string'
+      ? webpackConfigFactory
+      : webpackConfigFactory(env);
```
zamotany:
```js
typeof webpackConfigFactory === 'function'
  ? webpackConfigFactory(env)
  : webpackConfigFactory
```

src/utils/makeReactNativeConfig.js
jukben: Oh, damn, it should be an object. 😅

package.json
jukben: Damn, I've bet that Satya will write this. XD

src/compiler/worker/getConfig.js
```diff
@@ -14,10 +14,15 @@ module.exports = function getConfig(
...
+  let config;
+  try {
     // $FlowFixMe
-    require(configPath),
-    configOptions,
-    platform
-  );
+    config = require(configPath);
+  } catch (e) {
+    throw new Error(
```
satya164: We should default to the following config when none is present:

```js
import { createWebpackConfig } from 'haul';
 
export default {
  webpack: createWebpackConfig({
    entry: './index.js',
  }),
};
```

So we should do something like this:
1. Check if the config file is present, if not, use default config
2. If config file is present, try to load it

I'm not sure if we should be catching the error and throwing a new error because it'll shadow the original error which caused the `require` to fail, such as syntax error, which will make it harder to figure out how to fix the issue.

src/compiler/worker/getConfig.js
```diff
@@ -14,10 +14,15 @@ module.exports = function getConfig(
...
+      'Haul configuration cannot be loaded. Have you provided valid JS file?'
+    );
+  }
+
+  return makeReactNativeConfig(config, configOptions, platform);
```
satya164: This should do something like `config.__esModule ? config.default : config` so we support both ES exports and node exports.

src/compiler/worker/getConfig.js
```diff
@@ -14,10 +14,15 @@ module.exports = function getConfig(
...
 ) {
-  return makeReactNativeConfig(
+  let config;
+  try {
     // $FlowFixMe
-    require(configPath),
-    configOptions,
-    platform
-  );
+    config = require(configPath);
```
satya164: Does this work with the current code? The config should actually be `require(configPath).default` in this case since the export is ES export, and we're importing with node's `require`

fixtures/example-app/haul.config.js
```diff
@@ -0,0 +1,7 @@
+import { createWebpackConfig } from 'haul';
+
+export default {
+  webpack: createWebpackConfig(({ platform }) => ({
```
satya164: Unnecessary function syntax and unused argument

src/compiler/worker/getConfig.js
Krizzu: Also, should it worth pointing out that we fallback to default config?

src/utils/makeReactNativeConfig.js
```diff
@@ -224,13 +230,26 @@ function makeReactNativeConfig(
...
+  }
+
+  const webpackConfig =
+    typeof webpackConfigFactory === 'function'
+      ? webpackConfigFactory(env)
+      : webpackConfigFactory;
 
-  const config = Object.assign({}, defaultWebpackConfig, userConfig, {
-    entry: injectPolyfillIntoEntry(userConfig.entry, polyfillPath),
+  const config = merge(defaultWebpackConfig, webpackConfig, {
```
satya164: I don't think we should be deep merging webpack configs. This could produce weird results. Also the config received here should already a full webpack config and we shouldn't do any changes to that (see below).

src/utils/makeReactNativeConfig.js
```diff
@@ -270,4 +289,20 @@ function injectPolyfillIntoEntry(
...
-module.exports = { makeReactNativeConfig, injectPolyfillIntoEntry };
+function createWebpackConfig(
+  configBuilder: ConfigOptions => HaulConfig | HaulConfig
+) {
+  return (options: ConfigOptions) => {
+    if (typeof configBuilder === 'function') {
+      return configBuilder(options);
+    }
+
+    return configBuilder;
```
satya164: `createWebpackConfig` should return a valid webpack config with haul specific changes already done. Basically, most of the stuff from `makeReactNativeConfig` needs to be moved to this function.

src/compiler/worker/getConfig.js
satya164: @Krizzu we could show a log when starting haul. Something like: "No configuration file found. Using default configuration."

zamotany: Won't work properly, since it's used by worker, and the UI is a react app. The log won't be visible

src/compiler/worker/getConfig.js
zamotany: https://github.com/callstack/haul/blob/fde95a923d288b91c3fdeba301275a8692123b7d/src/utils/makeReactNativeConfig.js#L236-L237

src/compiler/worker/getConfig.js
zamotany: https://github.com/callstack/haul/blob/fde95a923d288b91c3fdeba301275a8692123b7d/src/utils/makeReactNativeConfig.js#L236-L237

src/utils/makeReactNativeConfig.js
zamotany: That doesn't make any damn sense.
> The createWebpackConfig utility can take some common options has return a webpack configuration function which takes the env argument. 

What common options? What are those? Either you need to be more specific in the proposal or this should be left as it is.

src/compiler/worker/getConfig.js
satya164: We should use the `__esModule` check for this  since that's what Babel does.

src/utils/makeReactNativeConfig.js
zamotany: Still doesn't make any sense when compared with proposal - once it returns a function once an object. It's inconsistent, so the best fallback is a function.

fixtures/example-app/haul.config.js
jukben: It's an example

src/compiler/worker/getConfig.js
jukben: Yeah, check for `__esModule` makes sense. It's definitely cleaner. 👍

src/compiler/worker/getConfig.js
jukben: Yeah, agree that we should fall back to the default configuration I wasn't aware that it's possible. But I like the idea. 👍

jukben:
I will reimplement this PR. Stay tuned. 👍

fixtures/example-app/haul.config.js
satya164: It's a fixture, but it's still unnecessary

src/utils/getConfig.js
```diff
@@ -0,0 +1,29 @@
...
+
+const { makeReactNativeConfig } = require('./makeReactNativeConfig');
+
+module.exports = function getConfig(
+  configPath: string,
+  configOptions: any,
+  platform: Platform
+) {
+  let config;
+  try {
```
satya164: Can you remove the try/catch?

> I'm not sure if we should be catching the error when `require` and throwing a new error because it'll shadow the original error which caused the require to fail, such as syntax error, which will make it harder to figure out how to fix the issue.

satya164:
We will also need to support the old config format as well for now (with a deprecation warning).

src/utils/getConfig.js
jukben: Yeah, I agree it would be probably better. I will remove it.

jukben:
@satya164 that actually make a sense. I will take a look on it.

src/utils/makeReactNativeConfig.js
```diff
@@ -237,6 +248,59 @@ function makeReactNativeConfig(
...
+   * 
+   * module.exports = {
+   *   entry: './index.js',
+   * };
+   * 
+   * module.exports = ({ platform }) => ({
+   *   entry: `./index.${platform}.js`,
+   * });
+   */
+  if (!Object.prototype.hasOwnProperty.call(userWebpackConfig, 'webpack')) {
```
jukben: Anyone has better idea how to check "old format" of the config file? @satya164 @zamotany

jukben:
@satya164 Please, re-check 👍

src/utils/makeReactNativeConfig.js
satya164: @jukben it'll be webpack.haul.js whereas new one is haul.config.js

jukben: That's only defaulted config, we support passing whatever file name user like. So check by a name file is definitely not a bullet-proof idea.

satya164: What do you mean by "we support passing whatever file name"? How can you specify a file with a different name for the config? I am not aware of a way, and the docs don't mention of any https://github.com/callstack/haul/blob/master/docs/Configuration.md

jukben: Yeah, you are able to `yarn haul start --config webpack.haul.js`

jukben: https://github.com/callstack/haul/blob/master/src/commands/start.js#L189 the same for bundle - and I like this feature, we should document it

satya164: Interesting. Ok so seems `const isLegacy = typeof config === 'function' || !config.webpack;` should be enough.

jukben: Why not to stick with `Object.prototype.hasOwnProperty.call(userWebpackConfig, 'webpack')`?

src/constants.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  DEFAULT_HAUL_CONFIG_FILE: 'haul.config.js',
```
satya164: Can you change the name to `DEFAULT_CONFIG_FILE_PATH`?

src/utils/makeReactNativeConfig.js
```diff
@@ -237,6 +248,62 @@ function makeReactNativeConfig(
...
+   * module.exports = ({ platform }) => ({
+   *   entry: `./index.${platform}.js`,
+   * });
+   */
+  const isLegacy =
+    typeof userWebpackConfig === 'function' || !userWebpackConfig.webpack;
+
+  if (isLegacy) {
+    logger.warn(
+      'You using deprecated style of the configuration. Please follow the docs to the upgrade.'
```
satya164: Can you add a link to the docs along with the message?

src/utils/makeReactNativeConfig.js
```diff
@@ -237,6 +248,62 @@ function makeReactNativeConfig(
...
+
+  const {
+    webpack: webpackConfigFactory /* , ...haulConfig */,
+  } = userWebpackConfig;
+
+  if (
+    typeof webpackConfigFactory !== 'function' &&
+    typeof webpackConfigFactory !== 'object'
+  ) {
+    throw new Error('You have provided a wrong Webpack configuration.');
```
satya164: Make the error message something like "The webpack configuration must be an object or a function returning an object" and maybe add a link to the configuration docs.

jukben:
> Are you handling the case where no config file is present? I couldn't find it in the PR. It should be something like this: 

No, I wasn't aware that we are on "no-config" hype train. 🚂 I will fix it.

package.json
```diff
@@ -5,6 +5,7 @@
   "bin": {
     "haul": "./bin/cli.js"
   },
+  "main": "src/index.js",
```
satya164: Is this related to this PR somehow?

BTW configure `eslint.autoFixOnSave` in your editor instead of prettier so you don't format JSON files accidentally.

fixtures/example-app/haul.config.js
```diff
@@ -0,0 +1,7 @@
+import { createWebpackConfig } from 'haul';
```
satya164: Maybe will be nice to delete this file in one of the fixtures to ensure our default config works properly

src/utils/makeReactNativeConfig.js
```diff
@@ -237,6 +265,74 @@ function makeReactNativeConfig(
...
+   * module.exports = ({ platform }) => ({
+   *   entry: `./index.${platform}.js`,
+   * });
+   */
+  const isLegacy =
+    typeof userWebpackConfig === 'function' || !userWebpackConfig.webpack;
+
+  if (isLegacy) {
+    logger.warn(
+      'You using a deprecated style of the configuration. Please follow the docs for the upgrade.'
```
satya164: Can we add a link to the docs?

src/utils/makeReactNativeConfig.js
```diff
@@ -237,6 +265,74 @@ function makeReactNativeConfig(
...
+  const {
+    webpack: webpackConfigFactory /* , ...haulConfig */,
+  } = userWebpackConfig;
+
+  if (
+    typeof webpackConfigFactory !== 'function' &&
+    typeof webpackConfigFactory !== 'object'
+  ) {
+    throw new Error(
+      'The webpack configuration must be an object or a function returning an object.'
```
satya164: Can we add a link to the docs?

package.json
jukben: It's we need to expose `createWebpackConfig` function.

src/utils/makeReactNativeConfig.js
jukben: Yeah, all documentation is still TODO, check the header of this PR.

src/utils/makeReactNativeConfig.js
jukben: Yeah, all documentation is still TODO, check the header of this PR.

fixtures/example-app/haul.config.js
jukben: Agree.

jukben:
Documentation updated. This should be ready now. I just have to wait for CI fix coming from #339. 💪 @satya164

src/utils/makeReactNativeConfig.js
jukben: Updated.

src/utils/makeReactNativeConfig.js
jukben: Updated.

src/resolvers/AssetResolver.js
```diff
@@ -14,7 +14,7 @@ type Request = {
 
 type Options = {
   test?: RegExp,
-  platform: 'ios' | 'android',
+  platform: string,
```
Krizzu: why change it to string?

src/utils/makeReactNativeConfig.js
```diff
@@ -270,4 +366,35 @@ function injectPolyfillIntoEntry(
   return userEntry;
 }
 
-module.exports = { makeReactNativeConfig, injectPolyfillIntoEntry };
+function createWebpackConfig(configBuilder: WebpackConfigFactory) {
+  return (options: EnvOptions) => {
+    const haulWebpackConfiguration =
+      typeof configBuilder === 'function'
```
Krizzu: Would it be okay to make one more check for `configBuilder`, if it's our required format? It's exposed to user directly and yet, we don't check if even `entry` is valid.

docs/Configuration.md
```diff
@@ -1,60 +1,66 @@
...
 
-The function will receive options as the first parameter, and the configuration used by `Haul` as the second parameter.
+The function will receive options as the first parameter.
 
 Options is an object with the following shape:
 
 ```js
 type Options = {
-  platform: 'ios' | 'android', // Current platform
+  platform: "ios" | "android", // Current platform
```
zamotany: Why??

docs/Configuration.md
```diff
@@ -1,60 +1,66 @@
 # Configuration
 
-The `webpack.haul.js` file is a webpack configuration used by `Haul`. The simplest config looks like this:
```
zamotany: You're markdown formatter has different config than our prettier config:
* we use single quotes instead of double quotes
* trailing spaces are added everywhere when it's compatible with ES5 (objects, arrays etc)

package.json
zamotany: @satya164 As @jukben said it is required from node to know there to looks for `createWebpackConfig`.

src/resolvers/AssetResolver.js
jukben: Yeah, I had some problems coming from Worker that Flow can't recognize enum in common string. ¯\_(ツ)_/¯

I'm not really satisfied with it, maybe we should investigate it once more. But it might be better to do it in separate PR. What do you think? Maybe I would use some help there.

src/compiler/Compiler.js
```diff
@@ -88,6 +88,10 @@ module.exports = class Compiler extends EventEmitter {
       });
     });
 
+    fork.on(Events.LOG, ({ message, logger }) => {
+      this.emit(Events.LOG, { platform, message, logger });
```
zamotany: `logger` should be named `type`

src/resolvers/AssetResolver.js
zamotany: Type for `platform` is defined here: https://github.com/callstack/haul/blob/8f84cbb077b2a17b8ae9a921ef32b4d590444810/src/types.js#L56

src/utils/makeReactNativeConfig.js
```diff
@@ -211,16 +228,27 @@ const getDefaultConfig = ({
...
+/** 
+ * Return React Native config
+ * 
+ * @deprecated
+*/
+function DEPRECATEDMakeReactNativeConfig(
+  userWebpackConfig: DEPRECATEDWebpackConfigFactory,
   options: ConfigOptions,
-  platform: 'ios' | 'android'
+  platform: string
```
zamotany: https://github.com/callstack/haul/blob/8f84cbb077b2a17b8ae9a921ef32b4d590444810/src/types.js#L56

docs/Configuration.md
jukben: I'm wondering where it's coming from because I'm using the normal prettier with eslint integration. Could you share your config?

src/utils/makeReactNativeConfig.js
jukben: Yeah, I had some problem with type inheritance. I will try once more. 

Flow claimed that he can't allow a string as an enum.

src/utils/makeReactNativeConfig.js
jukben: I don't think it's necessary, as we don't handle errors in config itself as well (for easier debug on user's side) we should trust user that he knows what he doing.

cc @satya164 @zamotany

satya164: Yes, it'll be good validate options passed to `createWebpackConfig`.

>  as we don't handle errors in config itself as well

yeah, but webpack will throw in that case, so it'll be duplicating the logic

> for easier debug on user's side) we should trust user that he knows what he doing

that was for `require` statement. when user returns their own webpack config, we should trust the user. but for arguments to our `createWebpackConfig`, if we don't validate it, it'll make debugging harder

jukben: I got it that it was for `require`. But this is the same case – we should trust the user. 

```import { createWebpackConfig } from 'haul';

export default {
  webpack: env => {
    const config = createWebpackConfig({
      entry: `./index.js`,
      include: /node_modules/,
    })(env);

    config.plugins.push(new CaseSensitivePathsPlugin());

    return config;
  },
}```

**Comment by jukben** [comment]
Hm. It looks that I have to resolve few conflicts.

File: src/utils/makeReactNativeConfig.js (hunk 20 above)
Comment (satya164) [reply src/utils/makeReactNativeConfig.js]: As I said, when user returns their own webpack config, we should trust the user. But @Krizzu is suggesting to validate arguments to `createWebpackConfig`, which is our own function. We know what are valid arguments and the user cannot pass arbitrary config here.

Comment (satya164) [reply src/utils/makeReactNativeConfig.js]: > In this example, configBuilder is an anonymous function

Comment (jukben) [reply src/utils/makeReactNativeConfig.js]: As I said, it has not to be, as I've shown you in the example literally one post above.

Comment (jukben) [reply src/utils/makeReactNativeConfig.js]: So If I understand you correctly, you proposing to check if `haulWebpackConfiguration` has `entry` field? Of if user passing to `createWebpackConfig` object or function with arity 1 (we accept one argument - env object)?

Comment (jukben) [reply src/utils/makeReactNativeConfig.js]: It might be a good idea to check if the object has keys what we are going to use - currently, we support only "entry". I can use Joi as object schema validator (https://github.com/hapijs/joi) to be prepared for rich cases. What do you think? @satya164

Comment (satya164) [reply src/utils/makeReactNativeConfig.js]:
```js

const config = typeof configBuilder === 'function' ? configBuilder(options) : configBuilder;

if (typeof config !== 'object' || config === null) {
  throw new Error(`The arguments passed to 'createWebpackConfig' must be an object or a function returning an object.`).
}

if (typeof config.entry !== 'string') {
  throw new Error(`The 'entry' property must be a string and point to your app's entry point (usually 'index.js').`);
}

if (!fs.existsSync(path.resolve(process.cwd(), config.entry))) {
  throw new Error(`The file '${config.entry}' doesn't exist. It should point to your app's entry point (usually 'index.js')`);
}
```

satya164: Don't think we need any validation library right now. If it gets more complex in future, we'll add one.

satya164:
@zamotany can you comment about avoiding mutating the resulting config which happens in your PR? it needs to be changed in this PR or yours.

src/utils/makeReactNativeConfig.js
jukben: at Joi – ok, makes sense.

src/resolvers/AssetResolver.js
rdy: If it is a string it would be potentially easier to extend haul to handle other platforms. I'd prefer to keep it this way to allow for easier extension.

rdy:
Would you be open to opening up the configuration for the haste resolver plugin as well? We're currently forking it to allow for custom platforms to be added with haul but we would need a hook in the config to use this approach instead of forking the whole thing. I'd specifically like to allow a set of modules in addition to `react-native` to be added.

jukben:
@rdy Interesting idea but I'm afraid I do not follow you completely. Could you please elaborate more how `providesModuleNodeModules` will avoid configuring plugins by your own `config.plugins.push`? Where will be those plugins defined and how?

rdy:
@jukben right now the providesModules equivalent in haul is injected with the `HasteResolver` plugin. You currently only get the choice of resolving `react-native`. It would be more convenient to have a configuration option that specifies this with `[ 'react-native' ]` as the default, but allow out of tree platforms like `react-native-windows` etc to inject their Haste modules.

This way if I wanted to add out of tree platforms to react-native I would only need to add:
```
providesModules: ['react-native', 'react-native-windows'] 
```
This is similar to the way that metro does it with rn-cli.

I guess what i'm asking for is, can we expose the HasteResolver plugin configuration in this new config object you are building?

satya164:
> can we expose the HasteResolver plugin configuration

We can add some parameters and export `HasteResolverPlugin`. Then you can use this plugin to add more haste stuff.

rdy:
@satya164 i totally agree, that a babel-plugin re-writing the haste modules to relative paths is a better long term solution. I may undertake that myself when I have a chance. But perhaps we could use the configuration option to remove the `HasteResolver` when you don't give it providesModules in that case. Right now there isn't any way to opt out of haste. If you are on a project that has totally adopted Haul and webpack you don't need haste to fix the platform imports so it would be faster to skip (if we somehow had a version of react-native itself that was haste free). I was thinking of using this configuration as a means to help those migrating from metro, who currently rely on haste.

thymikee:
@jukben mind merging `next` branch and continue working on this?

satya164:
You'll also need to address this comment: https://github.com/callstack/haul/pull/339#discussion_r171362226

satya164:
BTW you prolly need to do a rebase so that @zamotany's commits appear in the diff.

thymikee:
Yea, I've tried to ease things by merging, but that's just too hard to get right, since I was not in the topic (sorry about that). I've already talked to @jukben and he's going to revert the stuff I did and merge/rebase once again.

zamotany:
Good to go, lint and test are passing @thymikee 
@jukben Please fix flow errors.

There's also small problem, right now `You using a deprecated style of the configuration [...]` warning gets piped to `<stdout.log`. We need to change that.

docs/Configuration.md
```diff
@@ -1,60 +1,66 @@
...
+
+    config.plugins.push(new CaseSensitivePathsPlugin());
+
+    return config;
+  }
+};
 ```
 
-See the [Webpack website](https://webpack.js.org/) to learn more about configuring Webpack.
+With this approach wou can also take advantage from [webpack-merge](https://github.com/survivejs/webpack-merge). See the [Webpack website](https://webpack.js.org/) to learn more about configuring Webpack.
```
satya164: `wou`

satya164: advantage of*

integration_tests/utils.js
```diff
@@ -109,6 +110,25 @@ function replaceTestPath(string: string) {
   return string.replace(/^\W+(.*)integration_tests/gm, '<<REPLACED>>');
 }
 
+function fetchBundle(url: string): Promise<any> {
```
satya164: You don't seem to use this anywhere

jukben:
This is ready to go. Finally! 🎉 

Just wait for @thymikee. He will implement jest-diff to reduce size of snapshots. Which is a good idea. 🙏

thymikee:
I've left one complete snapshot inside `getConfig` and used diffs and asymmetric assertions for the rest

thymikee:
cc @satya164 one final review?

AleksandrZhukov:
@jukben What about updating docs? https://github.com/callstack/haul/blob/master/docs/Configuration.md

jukben:
@AleksandrZhukov Oh snap. Probably some flaw in the rebase. 😔 Can you bring it back in the PR? https://github.com/callstack/haul/blob/75bf51096909e8a7a3cd031060253240db3f3978/docs/Configuration.md

AleksandrZhukov:
@jukben thx

PR #377, opened 2018-03-01
gracefully add gradle cliPath config

### LINKED ISSUE #376: `haul init` clobbers existing project.ext.react gradle config
### Current Behavior
`haul init` adds the following block to my project's `android/app/build.gradle` above the `react.gradle` import, which overwrites my existing `project.ext.react` assignment.

```
project.ext.react = [
    cliPath: "node_modules/haul/bin/cli.js"
]
```

### Expected Behavior
`haul init` should merge the new `cliPath` property into `project.ext.react` config if it exists, *or* fall back to the current behavior and create a new `project.ext.react` block.

```
module.exports = () => ({
  entry: `./index.js`,
});
```

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-beta.13
| react-native     |  0.53.0
| node             | v9.3.0
| npm or yarn      | v1.3.2

Workaround is just to manually merge the two `project.ext.react` blocks, but it'd be nice if `haul init` did this for you.

It looks like it just needs a bit more logic where it adds this block [here](https://github.com/callstack/haul/blob/33775258ab1b801d1e0a4762a72fc3b78844c0bf/src/commands/init.js#L266)

I'm happy to PR a fix in the coming week : )

FLGMwt:
Fixes #376 

Instead adding a `project.ext.react` initialization with the `cliPath` for haul which could potentially clobber existing config, this initializes the `project.ext.react` config map if necessary, and then explicitly assigns the `cliPath` value.

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -250,11 +250,11 @@ const addToGradleBuild = async (cwd: string) => {
   let project = fs.readFileSync(entry).toString();
 
   const cliString = '"node_modules/haul/bin/cli.js"';
-
   const gradleConf = `
-    project.ext.react = [
-      cliPath: ${cliString}
-    ]
+    if (!project.hasProperty('react')) {
+      project.ext.react = [:]
+    }
+    project.ext.react['cliPath'] = ${cliString}
   `;
 
   // Are we already integrated?


thymikee:
Should we care about running `haul init` multiple times? Looks like it will just append this block. We should at least bail out with a logger message what happened.

Krizzu:
@thymikee do we do any checks for multiple `haul init` run? Maybe we should somehow store the information that this already happened?

thymikee:
For now I think just blindly regexing `project.ext.react` would work (and log there's already an entry, so bail out – that may give someone idea of why Haul is not updating their config)

jukben:
Is this ready? @FLGMwt Have you tested it with a fresh project? 🙏

FLGMwt:
I believe I tested it with both a fresh and a dirty project (one with an existing ext.react), but it's been a while, so I'll test again within the next two days.

jukben:
Thanks! Then we are ready to go. (just be sure that you use RN compatible with Haul <0.56).

thymikee:
Oh, hi @FLGMwt 😄

PR #381, opened 2018-03-05
Upgrade to Webpack 4

thymikee:
@czystyl can you make me a favor and test it once more? I've updated webpack to stable v4.1.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "source-map": "^0.5.6",
     "strip-ansi": "^3.0.1",
     "thread-loader": "^1.1.1",
-    "webpack": "3.8.1",
+    "webpack": "^4.0.0-beta.1",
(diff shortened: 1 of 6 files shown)


thymikee:
For now I've verified iOS works, with HMR too.
I get these deprecation warning though:
```
(...) Tapable.plugin is deprecated. Use new API on `.hooks` instead
(...) The callback argument was splitted into resolveContext and callback.
(...) The type arguments (string) is now a hook argument (Hook). Pass a reference to the hook instead.
```

czystyl:
It appeared just after the update

czystyl:
It looks like everything works except HRM with a Chrome debugger on Android

thymikee:
As a followup we'll migrate to new plugins API (ref: https://github.com/webpack/webpack/search?utf8=%E2%9C%93&q=Compiler.hooks&type=)

PR #388, opened 2018-03-12
Read rn-cli.config.js to add additional platforms

rdy:
When the rn-cli.config.js file is detected add the platforms specified to haul.

### DIFF
--- a/src/commands/start.js
+++ b/src/commands/start.js
@@ -16,7 +16,7 @@ const messages = require('../messages');
 const exec = require('../utils/exec');
 const getWebpackConfig = require('../utils/getWebpackConfig');
 const { isPortTaken, killProcess } = require('../utils/haulPortHandler');
-
+const { extraPlatformsDescriptions } = require('../utils/loadRnCli');
 const { makeReactNativeConfig } = require('../utils/makeReactNativeConfig');
 
 /**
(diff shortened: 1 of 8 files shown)


rdy:
I needed to base this on the commit prior to the webpack4 branch to keep it running with my sample application that is still using webpack3. I tested it earlier against the webpack4 one and there doesn't seem to be any problem with it.

satya164:
I think additional platforms should a be config option passed to `createWebpackConfig` https://github.com/callstack/haul/pull/372

Regarding reading `rn-cli.config.js`, why would we want to configure platforms in that instead of in haul's config? I think it'll be weird if we read and support some of the options from `rn-cli.config.js` while ignore others.

thymikee:
I agree with @satya164

rdy:
I just caught up on #372. That seems like a reasonable way to do it, but there is still some additional configuration that needs to be passed for the haste resolver plugin. If we can inject the modules for that in the configuration that would be cool. When is the plan for the new configuration to land? I'm happy to lend a voice to that side to make it easier for new platform targets to be added to haul.

src/commands/start.js
```diff
@@ -179,9 +179,10 @@ module.exports = ({
           value: 'android',
           description: 'Serves Android bundle',
         },
+        ...extraPlatformsDescriptions(),
         {
           value: 'all',
-          description: 'Serves both platforms',
+          description: 'Serves all platforms',
```
matthargett: 💯

thymikee:
Closing as https://github.com/callstack/haul/pull/372 is merged. We'll soon release RC with this change (you can track `next` branch)

PR #393, opened 2018-03-27
remove haste references from polyfillEnvironment

rdy:
Hi,

This is part of a greater change to get tree shaking to work with react-native. In order to use webpack's tree shaking I need to remove the use of haste. I'm planning to start a discussion upstream in react-native but I also wanted to make the haste references that haul uses in polyfillEnvironment.

I'm planning to use this fork to show how someone could use haul and webpack with react-native to get tree shaking to work on their react-native apps. In my example on an android test project is was able to shrink the bundle size from 500k with metro to 260k using haul.

I'm also open to writing a babel-plugin to de-haste the source but I thought I would try to make these small changes in haul.

### DIFF
--- a/src/utils/polyfillEnvironment.js
+++ b/src/utils/polyfillEnvironment.js
@@ -6,6 +6,7 @@
  *
  * This file is loaded as a part of user bundle
  */
+import { NativeModules } from 'react-native';
 
 /* eslint-disable import/no-extraneous-dependencies */
 require('../../vendor/polyfills/console.js')(global);
@@ -32,7 +33,7 @@ if (!global.self) {
   global.self = global; /* eslint-disable-line */
 }
 
-require('InitializeCore'); // eslint-disable-line import/no-unresolved
+require('react-native/Libraries/Core/InitializeCore');
 
 require('../hot/client/importScriptsPolyfill');
 
@@ -53,7 +54,7 @@ if (process.env.NODE_ENV !== 'production') {
     // from URL from which the bundle was loaded. When using iOS simulator/Android emulator
     // or Android device it will be `localhost:<port>` but when using real iOS device
     // it will be `<ip>.xip.io:<port>`.
-    const { scriptURL } = require('react-native').NativeModules.SourceCode; // eslint-disable-line import/no-unresolved
+    const { scriptURL } = NativeModules.SourceCode;
     if (scriptURL) {
       [protocol, , origin] = scriptURL.split('/');
     }


src/utils/polyfillEnvironment.js
```diff
@@ -32,11 +32,12 @@ if (!global.self) {
   global.self = global; /* eslint-disable-line */
 }
 
-require('InitializeCore'); // eslint-disable-line import/no-unresolved
+require('react-native/Libraries/Core/InitializeCore');
 
 require('../hot/client/importScriptsPolyfill');
 
-const scriptURL = require('react-native').NativeModules.SourceCode.scriptURL; // eslint-disable-line import/no-unresolved
+const scriptURL = require('react-native/Libraries/BatchedBridge/NativeModules')
```
thymikee: there's an import for Native modules, so you can do
```js
const { NativeModules } = require('react-native');
const { scriptURL } = NativeModules.SourceCode;
```

rdy: yea, there is a trick i'm doing to prevent the use of `react-native` main entry point. I'll show you in the code example I'm using but the tree shaking doesn't work properly when all modules are exported from this entry point. The getters that do the dynamic require break the tree shaking, so I use webpack to alias that. That is why i'm reaching into the `react-native` library code to find the NativeModules.

thymikee: I'm curious how you import regular modules like `View` or `Text`. 

It's worth noting, that `react-native` is official public API and paths like `react-native/Libraries/BatchedBridge/NativeModules` may change in future as it's an implementation detail – I'm a bit concerned about that.

rdy: Yea, I used a babel plugin that I found to rewrite module imports, I suppose I could force that to be used for haul as well so that it gets rewritten. That was I can use NativeModules in the way that you suggested.

The babel-plugin transforms the imports that look like this:
```
import { Text } from 'react-native
```

Transforming them to:
```
import Text from 'react-native/Libraries/Text/Text'
```

So I could change my commit to use that directly. I provide my own traversal which looks at `react-native-implementation.js` and finds all the exported symbols and the paths to the requires.

rdy:
I rebased with the suggestion of referring to NativeModules using the `main` for react-native. In order for the tree shaking to work we will need to make sure `polyfillEnvironment.js` gets modified by babel with the same plugin I use to rewrite the module imports to support tree shaking.

rdy:
Ok, I've put together an example using `create-react-native-app` ejected you can see it here:
https://github.com/rdy/tree-shaking-example/tree/tree-shaking. I've chosen to use the android platform since I had the build tools on my machine already. You can build the bundles from the package.json scripts `bundle:metro` and `bundle:haul`.

The master branch contains a version where I've optimized the bundle size with haul slightly by tweaking the babel output instead of relying on `react-native-preset`.
Metro bundle: 505K
Haul bundle: 538K

The `tree-shaking` branch shows the changes from that where I've used my fork of `react-native` that removes haste as well as haul that doesn't use haste anymore. I've provided a transform for `babel-plugin-transform-imports`. I'm basically looking at the original main for react-native and supplying the correct import path, this allows webpack to be able to tree shake out the dead code.

Haul bundle with tree shaking: 254K

I've created a series of PDF using `source-map-explorer` to illustrate that pieces of the bundle are actually being remove correctly by webpack.

In order to build a running android version you would need to run the gradle commands to build the native java code in the node modules. I didn't commit those to my branch of react-native. From my tests locally the shaken version ran just fine.

Anyways I hope this provides enough of a picture why I'm looking at stripping haste in both `haul` and `react-native`.

rdy:
I've created a corresponding issue in `react-native` here:
https://github.com/facebook/react-native/issues/18598

Hopefully it will get the discussion going again on removing haste from react-native which should make this approach possible.

src/utils/polyfillEnvironment.js
```diff
@@ -53,7 +53,9 @@ if (process.env.NODE_ENV !== 'production') {
     // from URL from which the bundle was loaded. When using iOS simulator/Android emulator
     // or Android device it will be `localhost:<port>` but when using real iOS device
     // it will be `<ip>.xip.io:<port>`.
-    const { scriptURL } = require('react-native').NativeModules.SourceCode; // eslint-disable-line import/no-unresolved
+    const {
+      NativeModules: { SourceCode: { scriptURL } },
```
satya164: I prefer the previous code. The nested destructuring is hard to read.

rdy:
I switched to using the import at the top so I could keep the destructuring code similar to the previous implementation, but it should allow my `babel-plugin-transform-imports` to safely change the import to `import NativeModules from 'react-native/Libraries/BatchedBridge/NativeModules'`. I believe that as long as users are using `babel-preset-react-native` they should automatically get `babel-plugin-transform-es2015-modules-commonjs` which should convert that import into a require for those who don't want tree shaking.

Issue #395, opened 2018-03-29
Cannot GET /debugger-ui/


goodmind:
### Current Behavior

Cannot GET /debugger-ui/
### Expected Behavior

Open React Native debugger
### Haul Configuration (webpack.haul.js)

https://gist.github.com/goodmind/2c5d5a3c83ab8c8f77330ce260e0014e
### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-beta.13
| react-native     |  0.54.3
| node             | 9.6.1
| npm or yarn      | yarn 1.5.1

Krizzu:
Hey @goodmind 
Do you mind explain a bit more about what's happening? What error you see?

z4o4z:
Hi @Krizzu I had the same issue and after debugging haul server I found the problem:
To send `debugger.html` to the browser uses the following code:

```
    switch (req.path) {
      /**
       * Request for the debugger frontend
       */
      case '/debugger-ui': {
        const readStream = fs.createReadStream(path.join(__dirname, '../assets/debugger.html'));
        res.writeHead(200, { 'Content-Type': 'text/html' });
        readStream.pipe(res);
        break;
      }
```

But sometimes Chrome(maybe not only Chrome) add to the end of the URL `/` and the `req.path` looks like '/debugger-ui/' and the `debugger.html` cannot send to the browser because switch-case can't find the valid path.

To resolve this problem I've just added some search params to the URL: `http://localhost:8081/debugger-ui?search`.

The other solution just adds one more `case`:
```
      case '/debugger-ui/':
      case '/debugger-ui': {
        const readStream = fs.createReadStream(path.join(__dirname, '../assets/debugger.html'));
        res.writeHead(200, { 'Content-Type': 'text/html' });
        readStream.pipe(res);
        break;
      }
```

Krizzu:
@z4o4z Hey, thanks for that
Feel free to open a PR.

daihieptn97:
Try **turning off the virtual server** and then running **react-native run ios** with ios or react-native run android with android
**It worked for me**

PR #398, opened 2018-04-04
fix cannot get debugger-ui page (#395)

z4o4z:
Fixed the problem when debugger-ui page not found.

### DIFF
--- a/src/server/middleware/devToolsMiddleware.js
+++ b/src/server/middleware/devToolsMiddleware.js
@@ -38,6 +38,7 @@ function devToolsMiddleware(debuggerProxy) {
       /**
        * Request for the debugger frontend
        */
+      case '/debugger-ui/':
       case '/debugger-ui': {
         const readStream = fs.createReadStream(
           path.join(__dirname, '../assets/debugger.html')


Krizzu:
@z4o4z Would it be better to strip `req.path` from any slash/backslashes and do `switch` based on that? In this approach, we could catch more cases ([like this one](https://github.com/callstack/haul/blob/master/src/server/middleware/devToolsMiddleware.js#L65))

z4o4z:
@Krizzu should I create middleware which will add to the request object `fixedPath` key, and replace all `request.path` with the `request.fixedPath`? I can't replace `req.path` because it is getter. Or should I just update only in this middleware? 

```
req.fixedPath = req.path.replace(/\/$/, '');
```

z4o4z:
@Krizzu what do you think about this commit?

Krizzu:
@z4o4z Hey, forgive my absence!
This looks great! Although IMO, this approach (importing util function to every other middleware where we handle path) could be replaced with a middleware that parses path for us, like you proposed. In this case, we wouldn't need to remember to apply utility function in future middlewares.

z4o4z:
@Krizzu Hi, I thought about the middleware but we can't change the `req.path` because `path` it is getter in the express and can't be changed. The solution to add a new key to `req`, but I don't know what name to use for this key(I thought about `req.fiexPath`). Any ideas for the key name?

Krizzu:
`parsedPath`, `fixedPath`, `cleanPath`, up to you :) @z4o4z

Krizzu:
👏 Great Job!

PR #399, opened 2018-04-04
fix: remove escape characters around $@

ndillon1:
Remove escape characters around $@ to allow additional parameters to be used within the haul integrate script


thymikee:
Thank you @ndillon1!

PR #404, opened 2018-04-14
Fix/deprecated loader context

hedgepigdaniel:
Two things:

 * Fix a crash with webpack 4.5.0 due to the use of the previously deprecated and noe presumeably removed `this.options` loader context variable
 * Minor upgrades to some dependencies to prevent peer dependency warnings when using webpack 4


thymikee:
IIRC we have this fixed on `next` branch, but @zamotany will tell you more about it :)

PR #406, opened 2018-04-19
Upgrade dependencies to fix warnings and issues

hedgepigdaniel:
* `babel-loader` and `thread-loader` are upgraded to fix peer dependency warnings with webpack 4.0
* `babel-preset-react-native` is updated to fix peer dependency warnings in projects using haul, which get the latest version 4.0 by default.


thymikee:
Thanks for sending the PR! 🙂

PR #408, opened 2018-04-22
Seek for the deprecated config file 🔎

jukben:
I think it's a good idea to search for `webpack.haul.js` which is deprecated config name in favor of `haul.config.js`.

If no config is specified by `--config` Haul will try to find `haul.config.js` first, then it tries to find `webpack.haul.js` if no of them is found he will run in zero-configuration™ mode.

If a user specifies `--config` and that path is non-valid, Haul will throw an error says that.

@satya164 if you are reading this, feel free to rephrase the comments. 💪

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,4 +1,5 @@
 module.exports = {
+  DEPRECATED_DEFAULT_CONFIG_FILE_PATH: 'webpack.haul.js',
   DEFAULT_CONFIG_FILE_PATH: 'haul.config.js',
   DEFAULT_PORT: 8081,
 };
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,5 +1,5 @@
 module.exports = {
-  DEPRECATED_DEFAULT_CONFIG_FILE_PATH: 'webpack.haul.js',
-  DEFAULT_CONFIG_FILE_PATH: 'haul.config.js',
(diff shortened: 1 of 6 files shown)


satya164:
Currently we allow `webpack.haul.js` to contain new config format and `haul.config.js` to contain old config format, Maybe if `--config` option is not specified, and we find `webpack.haul.js`, we shouldn't allow it to contain new format and vice-versa.

jukben:
I see your concerns but I think it's not necessary. This solution works well for its purpose which is easy migration between `Haul` and `Haul/next`.

thymikee:
>[if] we find webpack.haul.js, we shouldn't allow it to contain new format and vice-versa.

Agreed. But for now I'm good with this solution. We can iterate later

src/constants.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  DEPRECATED_DEFAULT_CONFIG_FILE_PATH: 'webpack.haul.js',
   DEFAULT_CONFIG_FILE_PATH: 'haul.config.js',
```
thymikee: Future improvement: change these constants to `[DEPRECATED_]DEFAULT_CONFIG_FILENAME`, as it's not a path and it's actually a bit misleading while reading code in `getWebpackConfigPath.js`

jukben: Agree. I'm going to change it right now.

thymikee:
Feel free to merge once CI is green

PR #409, opened 2018-04-22
Better browser luncher 🌎

jukben:
This PR fixes uncaught exception when Haul tries to open Google Chrome and fails.

I'm using Google Chrome Canary as my primary browser and I don't have a Google Chrome even installed. I introduced new dependency which based on the platform tries run command. If all attempts are unsuccessful, Haul print a warning about it.


thymikee:
Damn, @jukben, you're hungry 😛

jukben:
Wait with the merge, I noticed that `opn` is async. I need to fix a code a little bit. 🤦‍♂️

thymikee:
Coolio

PR #412, opened 2018-04-22
Transition to babel 7

Krizzu:
So far, updated dependencies. Haul starts, but need to work on actual bundling

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,14 +31,14 @@
     "react-native": "*"
   },
   "dependencies": {
+    "@babel/core": "^7.0.0-beta.44",
+    "@babel/plugin-proposal-object-rest-spread": "^7.0.0-beta.44",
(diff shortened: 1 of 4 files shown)


thymikee:
This makes us >=0.56 compatible :<

Krizzu:
Fun is over

thymikee:
We can still use Babel 7 to compile Haul itself ;) https://github.com/callstack/haul/issues/416

Un3qual:
Any ETA or status updates on babel 7 compatability?

thymikee:
We don't have ETAs but we're happy to accept PRs :)
We need to put babel-core into peer dependencies and ask users to install a transition package `babel-core@^7.0.0-0`. There's a chance it will just work with such setup. As a tool/library we cannot fully transition to Babel 7, and neet this compat layer

Issue #418, opened 2018-04-30
allow out-of-tree platforms to be dynamically added


matthargett:
### Current Behavior
External platforms, such as Windows, cannot use haul without forking it. This is due the platforms array ['android','ios'] being hardcoded in makeReactNativeConfig.

### Expected Behavior
Out-of-tree platforms like Windows, Linux, and MacOS, should be able to use haul without forking haul.
In our initial rn-cli.config PR, we were told the new configuration would allow for out-of-tree platforms, but it turned out not to.

cc @grabbou ,we talked about this back in December and you had a clear idea on how to enable this.

jukben:
This could be closed, am I right, @acoates-ms? 💪

PR #171, opened 2017-06-14
Fix findProvidesModule throwing error in while running start command

zamotany:
### Current Behavior
1. `react-native init testApp`
2. `cd testApp && yarn add haul`
3. `yarn run haul -- --platform ios`
4. `react-native run-ios` (in other tab)
5. Stop `haul` by Ctrl+C and run again: `yarn run haul -- --platform ios`
```
 ERROR  Error running yarn run haul -- --platform ios

Error: ENOENT: no such file or directory, stat '/Users/paweltrysla/Documents/TrueCar/testRNApp/node_modules/react-native/third-party/glog-0.3.4/test-driver'
at Object.fs.statSync (fs.js:968:11)
at walk (/Users/paweltrysla/Documents/TrueCar/testRNApp/node_modules/haul/src/utils/findProvidesModule.js:60:21)
at fs.readdirSync.forEach.file (/Users/paweltrysla/Documents/TrueCar/testRNApp/node_modules/haul/src/utils/findProvidesModule.js:67:9)
at Array.forEach (native)
at walk (/Users/paweltrysla/Documents/TrueCar/testRNApp/node_modules/haul/src/utils/findProvidesModule.js:63:27)
at fs.readdirSync.forEach.file (/Users/paweltrysla/Documents/TrueCar/testRNApp/node_modules/haul/src/utils/findProvidesModule.js:67:9)
at Array.forEach (native)
at walk (/Users/paweltrysla/Documents/TrueCar/testRNApp/node_modules/haul/src/utils/findProvidesModule.js:63:27)
at fs.readdirSync.forEach.file (/Users/paweltrysla/Documents/TrueCar/testRNApp/node_modules/haul/src/utils/findProvidesModule.js:67:9)
at Array.forEach (native)
```

### Expected Behavior
No error
### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  `1.0.0-beta.0`
| react-native     |  `0.45.1`
| node             | `7.10.0`
| npm or yarn      | `yarn 0.23.4`

After first run of `react-native run-ios` in freshly created project, `/Users/paweltrysla/Documents/TrueCar/testRNApp/node_modules/react-native/third-party/glog-0.3.4/test-driver` is replaced with symlink, thus `fs.statSync` (https://github.com/callstack-io/haul/blob/master/src/utils/findProvidesModule.js#L60) throws error. Quick fix is to replace it with `fs.lstatSync` which handles symbolic links correctly.

PR #419, opened 2018-05-01
allow out-of-tree platforms to be dynamically added

acoates-ms:
This expands the config returned in haul.config.js to allow it to specify additional platforms, and expands the options used in createWebpackConfig to allow additional packages and platform extensions to be included in the haste map.

See #418

### DIFF
--- a/src/compiler/Fork.js
+++ b/src/compiler/Fork.js
@@ -42,7 +42,7 @@ module.exports = class Fork extends EventEmitter {
       // WebSocket connection is established after the Fork is created.
       transportServer.on('connection', socket => {
         const platformMatch = socket.upgradeReq.url.match(
-          /platform=(.*)([&]|$)/
+          /platform=([^&]*)/
         );
 
         if (!platformMatch) {
--- a/src/compiler/Fork.js
+++ b/src/compiler/Fork.js
@@ -41,9 +41,7 @@ module.exports = class Fork extends EventEmitter {
 
       // WebSocket connection is established after the Fork is created.
       transportServer.on('connection', socket => {
-        const platformMatch = socket.upgradeReq.url.match(
-          /platform=([^&]*)/
-        );
+        const platformMatch = socket.upgradeReq.url.match(/platform=([^&]*)/);
 
         if (!platformMatch) {
           throw new Error('Incorrect platform');
--- a/src/compiler/Fork.js
+++ b/src/compiler/Fork.js
@@ -42,7 +42,7 @@ module.exports = class Fork extends EventEmitter {
       // WebSocket connection is established after the Fork is created.
       transportServer.on('connection', socket => {
         const platformMatch = socket.upgradeReq.url.match(
-          /platform=(ios|android)/
+          /platform=(.*)([&]|$)/
         );
 
         if (!platformMatch) {
--- a/src/server/ui.js
+++ b/src/server/ui.js
@@ -156,6 +156,7 @@ class ServerUI extends React.Component<ServerPropsType, ServerStateType> {
     this.props.compiler.on(
       Compiler.Events.BUILD_PROGRESS,
       ({ progress, platform }) => {
+        this._ensurePlatformState(platform);
         this.setState(state => ({
           compilationProgress: {
             ...state.compilationProgress,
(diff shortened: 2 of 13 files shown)


src/utils/findProvidesModule.js
```diff
@@ -84,9 +84,11 @@ function findProvidesModule(directories, opts = {}) {
 
       // Throw when duplicated modules are provided from a different
       // fileName
+      /*
```
acoates-ms: Really the module map should be keeping track of (moduleName, platform) => file.  Since there is no rule that any foo.android.js and foo.ios.js file have to be in the same directory.  

But since platform isn't obviously available at the lookup time, I'm just doing last write wins here.  What this means practically is that you have to specify ['react-native', 'react-native-windows'] as you providesModuleNodeModules, so that the react-native-windows modules stomp on the react-native ones.  This also means that you have to only specify the react-native-windows package when building the windows platform, and not when building the ios/android platforms.

src/utils/findProvidesModule.js
```diff
@@ -57,7 +57,7 @@ function findProvidesModule(directories, opts = {}) {
   const modulesMap = {};
 
   const walk = dir => {
-    const stat = fs.lstatSync(dir);
+    const stat = fs.statSync(dir);
```
acoates-ms: This fixes an issue with having symlinks within the react-native package, which some package managers end up doing in some cases.

zamotany: No sure about this change: https://github.com/callstack/haul/pull/171

thymikee: Looks like this was a bug fixed in RN 0.47: https://github.com/facebook/react-native/commit/7d1981ef8245cd6267823a970054cbebb2d2fd14

jukben:
Looks good to me (besides `lstatSync` discussion)! Quality PR. 👏 I haven't time to test it, though I'm on vacation. Next week I'll find some time for it if someone wouldn't do it for me. 👍

thymikee:
@jukben mind looking at this once again?

src/compiler/Fork.js
```diff
@@ -42,7 +42,7 @@ module.exports = class Fork extends EventEmitter {
       // WebSocket connection is established after the Fork is created.
       transportServer.on('connection', socket => {
         const platformMatch = socket.upgradeReq.url.match(
-          /platform=(ios|android)/
+          /platform=(.*)([&]|$)/
```
thymikee: We pass only one platform to the Fork. Maybe we can just base this regex on `platform` param? wdyt @zamotany?

src/server/ui.js
```diff
@@ -330,8 +349,15 @@ class ServerUI extends React.Component<ServerPropsType, ServerStateType> {
...
         <Text style={styles.progressContainer}>
-          {this._makeProgressBar('ios', 'iOS', 4)}
-          {this._makeProgressBar('android', 'Android')}
+          {Object.keys(this.state.compilationProgress)
+            .sort()
+            .map(platform => {
+              return this._makeProgressBar(
+                platform,
+                platform,
+                7 - platform.length > 0 ? 7 - platform.length : 0
```
thymikee: Future reference: we'll need to make this more robust in case we have a platform longer than 7 characters :p

acoates-ms: If the platform is longer than 7 it wont die, it just wont align the progress bars.  -- We could push that logic up the tree, to have state to calculate the longest platform name and pass it down the tree.  I didn't think it was worth it right now since I don't think any of the major platforms are > 7 chars.

thymikee: Yea, this is rather something for a followup if someone comes up with a weird use case ;).

src/compiler/Fork.js
zamotany: What's the reasoning behind this change??

jukben: I see @acoates-ms wouldn't `([^&]*)` do the same job?

acoates-ms: Sure thing -- I've gone ahead and updated it.

zamotany: Alternatively we can make it more strict: `platform=([a-zA-z]+)`

acoates-ms: I don't think we should be that opinionated about what a valid platform name can be.  Ex: should '-' be a valid platform name character?  '_'?

thymikee:
Are we good merging this? :)

danforstinger-youi:
any idea which release this will make it in to or when this will be available for general consumption?

jukben:
@thymikee could you please release new RC? 🙏

danforstinger-youi:
Yes! That would be awesome. We are eager to consume this feature.

PR #422, opened 2018-05-03
Implement trivial delta bundle support (Fixes #334)

### LINKED ISSUE #334: React Native 0.52 Android Delta Bundles
In the latest version of react native, android by default tries to load a new 'delta' bundle.
(See: https://github.com/facebook/react-native/blob/79e24ede40b2508aaa77b8ff3876d3dbf4cfe6d8/ReactAndroid/src/main/res/devsupport/xml/preferences.xml)
This results in an error when using Haul because the app requests an `index.delta` file that is not being served.

The fix is simple though, open the dev menu, select the "Dev Settings" option, then uncheck the "Use JS Deltas" box.  Close then re-open your app and it should request the normal javascript bundles as before.

simonbuchan:
This doesn't get any actual benefits of using delta bundles, it simply
repackages the current bundle in a trivial delta bundle. This seems to
have worse performance that using the standard format, so I added a
one-time warning (bike-shedding welcome on the wording)

This is still helpful in the case of automated tests which (re)install
and run the application immediately without user interaction, where
there is no chance to disable delta bundles.

If it helps, however, the format (as reverse-engineered from the
android client code) is described in case full delta support is added
later, or someone wants to know what the heck this is trying to do :)

Note that I also made the adb reverse call one-time, since it's pretty
spammy when reloading a lot!


Krizzu:
Hey @simonbuchan, great work. Your explanation for dbundles is pretty straight forward. I can imagine in the future we'll make a better use for them, but for now, it's superb.

Thanks.

simonbuchan:
An easy win would be adding the asset hash as the id, then serving `{}` if it hasn't changed for example!

I'm not very clear on why delta packages work the way they do unless FB just didn't want to keep the transformed output around in the packager? I was expecting some sort of hot reload type magic. As the comment mentions, perhaps the raw transfer to the device is a limiting factor?

It was getting pretty late last night, but I would like to do some more careful performance comparisons - our real app is really flakey under dev load even with RN packager so I'm a little suspicious of the feeling I got of it taking noticably longer with deltas.

And I assume docs?

hedgepigdaniel:
I often debug over a network in order to test on a physical device without being stuck to a computer or needing another USB port. I would imagine in that case that the transfer time could be a limiting factor, depending on the device/network.

simonbuchan:
Sure thing 👌, though it will be most of a day (being midnight here) before I get a chance

simonbuchan:
The semantic conflicts are from #419 allowing any platform, so if @acoates-ms wants to, feel free to double-check I didn't do something stupid. Things that stuck out were:
- The `Platform` type wasn't updated to `string` - I made this change and everything still seems to typecheck so 🤞
- I used a stricter regex for grabbing the platform out of the request name, matching what RN used. (it will still grab `weird/platform` as the platform out of `/not.a.weird/platform.bundle`, but hey)
- Fork.js should probably have been changed to use `req.query.platform` instead of using a regex, in case somebody uses a platform name that needs url-encoding (e.g. `dot-net`), but I wasn't sure if that was available, so I held off making the change.

Also the tests are broken on windows now (seems to just be the expected path asserts and invoking npm binstubs) - I've fixed those sorts of issues for detox, but I didn't want to poison this PR with them.

acoates-ms:
Looks ok.

src/server/util/getRequestBundleData.js
```diff
@@ -0,0 +1,68 @@
...
+      return null;
+    }
+
+    platform = match[1];
+  }
+
+  // Convert to a filename the compiler will find. This actually depends
+  // on the output configuration, so really this should be just picking it
+  // off the configuration for the provided platform.
+  // Also note the cast due to https://github.com/facebook/flow/issues/3554
```
thymikee: Can you explicitly use `// $FlowFixMe – https://github.com/facebook/flow/issues/3554` so it will error when this is fixed upstream?

simonbuchan: Sure. While I was there, I unpacked what I wrote in that comment above it and realized it was a bit nonsense and updated it - the actual thing to do to get rid of this is to change the webpack config to have `output.filename` set to `[path]/[name].bundle`, e.g. `src/entry.bundle` for `src/entry.js` to match what metro is doing. Obviously I don't really want to make that change here, since it probably breaks everything.

thymikee:
ping @zamotany

javajolt:
Any chance we might be able to get this feature in an upcoming release? The branch fixes our Android issues. Thanks!

thymikee:
Let's give it a try ;)

javajolt:
Thanks, @thymikee !

PR #434, opened 2018-05-22
Don't use deflate & better socket error handle 

### LINKED ISSUE #429: Haul crashes after upgrading from beta to 1.0-rc (zlib related)
After compilation success, haul server crashes with:

```
haul/src/server/index.js:58                                                                                                                                   
    throw err;
    ^

Error: invalid block type
    at InflateRaw.zlibOnError (zlib.js:153:15)
```

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  haul@1.0.0-rc.0
| react-native     |  react-native@0.55.3
| node             | v8.11.1
| npm or yarn      | 5.8.0

jukben:
I've discovered that we are using ws@2.2.0 which has the deflate option turned *on* by default. This issue is not actually about deflate but about `maxPayload`. (maxPayload error blown the zlib).

This is supposed to be just a walk around because I'm still not sure how the size is computed thought.

- My test project should be 42 633 230 byte. _(this is in Haul's logs or when I dowload the bundle into file via `wget -O bundle http://localhost:8081/index.ios.bundle` or something)_
- Default `maxPayload` ws@2.2.0 is 104 857 600 byte. (https://github.com/websockets/ws/blob/2.2.0/lib/WebSocketServer.js#L46)
- However, the `ws` claim that the bundle file is 144 242 145 byte (I had loged this in the context of very ws itself: https://github.com/websockets/ws/blob/2.2.0/lib/Receiver.js#L467).

I have also created new branch where I'm going to upgrade ws to latest version https://github.com/callstack/haul/tree/jukben/ws-upgrade

For context: https://github.com/websockets/ws/issues/1136

Not sure if this issue could be related to [memory-fs](https://github.com/webpack/memory-fs) what we use.

CLOSE #429

### DIFF
--- a/src/compiler/Compiler.js
+++ b/src/compiler/Compiler.js
@@ -37,6 +37,8 @@ module.exports = class Compiler extends EventEmitter {
         this.forks[platform] = this.initFork({ platform, options });
       }
 
+      if (!this.forks[platform]) return;
+
       // If the fork is compiling the bundle, attach listener to emit `REQUEST_FILE` once
       // the bundle is created, otherwise simply request the file. Callback will be then invoked in
       // `FILE_RECEIVED` listener.
--- a/src/compiler/Compiler.js
+++ b/src/compiler/Compiler.js
@@ -92,7 +92,13 @@ module.exports = class Compiler extends EventEmitter {
    * Create fork process and attach necessary event listeners.
    */
   initFork({ platform, options }: { platform: Platform, options: * }) {
-    const fork = new Fork({ platform, options });
+    let fork;
+    try {
+      fork = new Fork({ platform, options });
+    } catch (err) {
+      this.emit(Events.BUILD_FAILED, { platform, message: err.toString() });
+      return null;
+    }
 
     fork.on(Events.FILE_NOT_FOUND, ({ taskId }) => {
       const { callback, awaitingCount } = this.tasks.pop(taskId);
--- a/src/compiler/Compiler.js
+++ b/src/compiler/Compiler.js
@@ -95,8 +95,8 @@ module.exports = class Compiler extends EventEmitter {
     let fork;
     try {
       fork = new Fork({ platform, options });
-    } catch (err) {
-      this.emit(Events.BUILD_FAILED, { platform, message: err.toString() });
+    } catch (message) {
+      this.emit(Events.BUILD_FAILED, { platform, message });
       return null;
     }
 
--- a/src/compiler/Fork.js
+++ b/src/compiler/Fork.js
@@ -1,7 +1,7 @@
 /**
  * Copyright 2017-present, Callstack.
  * All rights reserved.
- * 
+ *
  * @flow
  */
 
@@ -50,6 +50,13 @@ module.exports = class Fork extends EventEmitter {
         }
 
         forks[platformMatch[1]].setSocket(socket);
+
+        socket.on('error', err => {
+          this.emit(Events.BUILD_FAILED, {
+            message: `Socket: ${err.toString()}`,
+          });
+          throw err;
+        });
       });
     }
 
--- a/src/compiler/Fork.js
+++ b/src/compiler/Fork.js
@@ -53,7 +53,7 @@ module.exports = class Fork extends EventEmitter {
 
         socket.on('error', err => {
           this.emit(Events.BUILD_FAILED, {
-            message: `Socket: ${err.toString()}`,
+            message: `Socket: ${err}`,
           });
           throw err;
         });
--- a/src/compiler/createWebSocketServer.js
+++ b/src/compiler/createWebSocketServer.js
@@ -39,7 +39,11 @@ module.exports = function createWebSocketServer() {
   const socketAddress = getSocketAddress();
 
   const httpServer = http.createServer();
-  const webSocketServer = new Server({ server: httpServer });
+  const webSocketServer = new Server({
+    server: httpServer,
+    perMessageDeflate: false,
+    maxPayload: 200 * 1024 * 1024,
+  });
 
   httpServer.listen(socketAddress);
 


thymikee:
Can we upgrade the `ws` lib by the way?

jukben:
@thymikee 
>I have also created new branch where I'm going to upgrade ws to latest version https://github.com/callstack/haul/tree/jukben/ws-upgrade

I will do it as a follow up. 👍

jukben:
The "zlib error" was just a hidden maxPayload issue. The deflate was malfunctional because message reached its boundaries. Still trying to understand we have reached the limit in this case though.

thymikee:
So we could stay with the default settings for now

jukben:
@thymikee OK, I will turn it on again. However, I'd rather see a reason why should I something turn on than otherwise. Then had probably a good reason to turn it off in a higher version. 😁

jukben:
Regarding the size. I did some test using socat and I realized something weird. Then I've discovered that memory-fs what we are using returns Buffer by default. 

I've tried to do https://github.com/callstack/haul/blob/next/src/compiler/worker/initWorker.js#L85
```js
file: fs.readFileSync(filename).toString('utf8'),
```
The result was outstanding

```
138M    socket-sent-darwin-buffer.txt
44M     socket-sent-darwin.txt
```

So this is more question to @zamotany I guess. Do we have to send data as serialized Buffer? We can save lot of comunication and it works still like a charm. It could be broken with different charcode though.

lpinca:
@jukben if with charcode you mean encoding, keep in mind that text frames in the WebSocket protocol must be valid UTF-8 encoded.

Since you already have a buffer, what you can do is encoding metada in a binary format and concatenate it with the file buffer. On the receiving end, you slice the buffer to get back the metadata and the file. It's more work and I don't know if it's viable for you but it's way more efficient as you have no UTF-8 validation and no direct calls to `buffer.toString()` (event though you will probably have to call `buffer.toString()` yourself, later).

jukben:
@lpinca Yeah I meant encoding. If the WebSocket protocol has to be valid UTF-8 it's even better, thanks for the clarification. 👍 

Thanks for the tip. I'm actually not sure if the `buffer.toString()` will cost us in this case so much that it makes a sense for us to optimize it in this way. Anyway thanks for the tip. Let's see what others say. I might be wrong.

satya164:
> @jukben: Do we have to send data as serialized Buffer

Yeah, otherwise it messes up encoding which is especially an issue with binary assets.

> @jukben: We can save a lot of communication

I'm not familiar with everything but correct me if I'm wrong. Does this really matter? Since it's all local, how slow can it be?

> @lpinca: using fs.readFileSync(filename, { encoding: 'utf8' }) to read the file

It'll probably break encoding for non-string files.

jukben:
@satya164 Yeah, the images (and other binary files) has to use Buffer, I was speaking about Bundle. Because the bundle was an issue in this case. But we can go just with bigger `maxPayload` and we are fine.

satya164:
> I was speaking about bundle JS assets

Don't think special casing text files is worth it if it doesn't provide any benefits. The question is, does it really improve the perf significantly to justify the additional complexity?

jukben:
@satya164 Probably not. I just wanted to know why we need such a big `maxPayload` (200 MB) when my bundle.js is ~40 MB. I'm totally fine to go just with this PR. 👍

src/compiler/Compiler.js
```diff
@@ -92,7 +92,13 @@ module.exports = class Compiler extends EventEmitter {
    * Create fork process and attach necessary event listeners.
    */
   initFork({ platform, options }: { platform: Platform, options: * }) {
-    const fork = new Fork({ platform, options });
+    let fork;
+    try {
+      fork = new Fork({ platform, options });
+    } catch (err) {
+      this.emit(Events.BUILD_FAILED, { platform, message: err.toString() });
+      return null;
```
satya164: Looks like returning `null` here will break code at line 43?

image

It checks the `.isProcessing` property. Maybe it's better to move emitting event to the callsite (line 37)?

src/compiler/Fork.js
```diff
@@ -50,6 +50,13 @@ module.exports = class Fork extends EventEmitter {
         }
 
         forks[platformMatch[1]].setSocket(socket);
+
+        socket.on('error', err => {
+          this.emit(Events.BUILD_FAILED, {
+            message: `Socket: ${err.toString()}`,
+          });
+          throw err;
```
satya164: Don't think anybody can catch this error. Should we throw it?

src/compiler/Compiler.js
```diff
@@ -92,7 +92,13 @@ module.exports = class Compiler extends EventEmitter {
    * Create fork process and attach necessary event listeners.
    */
   initFork({ platform, options }: { platform: Platform, options: * }) {
-    const fork = new Fork({ platform, options });
+    let fork;
+    try {
+      fork = new Fork({ platform, options });
+    } catch (err) {
+      this.emit(Events.BUILD_FAILED, { platform, message: err.toString() });
```
satya164: `toString` is probably unnecessary and should be done automatically when sending to worker.

lpinca:
One last thing then I'm out, be careful when working with such big strings. The maximum string length in V8 was ~256 MiB which is not too far from what you have now. The limit has been increased in newer versions of V8 but if I'm not wrong it's still 256 MiB in Node.js 6. An error is thrown if you try to create a string bigger than that.

zamotany:
@jukben as @satya164 said, we were sending buffer because of mime types issue, previously we actully head `fs.readFileSync().toString()`, but now we do it here: https://github.com/callstack/haul/blob/423c761f5a2736c59d860ed2f635e8c4ffe79d4f/src/server/middleware/compilerMiddleware.js#L46

That being said since V8 has a limit of 256, maybe we should split the file and send it in chunks. It shouldn't be difficult since we have this abstraction of `Compiler` and `Fork`.

src/compiler/Fork.js
jukben: It will be written in `.artifacts`

src/compiler/Compiler.js
jukben: I would rather be explicit.

src/compiler/Compiler.js
jukben: Agree. This going to crash the whole thing. With nice error and ⚠️icon next to the platform.

src/compiler/Compiler.js
satya164: It's pointless. the whole thing will be stringified in the end.

jukben: `toString()` will resolve to `message` if the throwned object is an instance of Error. `message` could break the code if someone throw something different. But fair point about stringified I'm going to change it.

PR #435, opened 2018-05-24
Run adb reverse when platform is 'all'

grahamlyus:
`adb reverse` is only run when `platform` is `android`, it should be run if `platform` is `all` also.


thymikee:
@zamotany can you please cherry-pick this onto `next` branch?

Krizzu:
@thymikee I think there is no need for that

PR #436, opened 2018-05-24
Android assets

### LINKED ISSUE #367: assetLoader android path issue when bundling
### Current Behavior
Importing images from the root folder or from a relative path outside of the root folder does not work when bundling for android. This works in a project created with `react-native init`

I've created a repo (https://github.com/grahamlyus/ImageHaul) that reproduces the issue (make sure the haul server is not running, otherwise assets load from localhost:8081).

Running `yarn haul bundle --platform android --dev true` from the `mobile` directory produces:
```
dist
├── drawable-mdpi
│   ├── _common_relative.jpg
│   ├── _root.jpg
│   └── images_child.jpg
└── index.android.bundle

1 directory, 4 files
```
Note that the root and relative images have an underscore prepended.

The output should be:
```
dist
├── drawable-mdpi
│   ├── common_relative.jpg
│   ├── images_child.jpg
│   └── root.jpg
└── index.android.bundle

1 directory, 4 files
```

### Expected Behavior
Importing images from anywhere configured by webpack should work in all modes.

Replacing the construction of `longname` as below solves the problem by removing the leading `../` and not adding a leading underscore if the asset is in the root folder:
```
  const normalizedUrl = url.replace(new RegExp(`^[\\.\\${path.sep}]+`), '')
  const normalizedName = normalizedUrl.length === 0 ? filename : `${normalizedUrl.replace(pathSepPattern, '_')}_${filename}`
  const longname = `${normalizedName.toLowerCase().replace(/[^a-z0-9_]/g, '')}.${type}`;
```

I can create a pull request if necessary.

I added resolve.modules to add a common folder (and full path to node_modules is also required when importing from common).
```
import path from 'path'

  ...defaults,
  entry: `./index.js`,
  resolve: {
    ...defaults.resolve,
    modules: [
      path.resolve(__dirname, '../common'),
      path.resolve(__dirname, './node_modules')
    ]
  }
});
```
### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-beta.13
| react-native     |  0.53.0
| node             |  8.9.1
| npm or yarn      | 1.3.2

grahamlyus:
Fix #367

### DIFF
--- a/src/loaders/assetLoader.js
+++ b/src/loaders/assetLoader.js
@@ -41,7 +41,12 @@ module.exports = async function assetLoader() {
   const assets = path.join('assets', config.bundle ? '' : config.platform);
   const suffix = `(@\\d+(\\.\\d+)?x)?(\\.(${config.platform}|native))?\\.${type}$`;
   const filename = path.basename(filepath).replace(new RegExp(suffix), '');
-  const longname = `${`${url.replace(pathSepPattern, '_')}_${filename}`
+  const normalizedUrl = url.replace(new RegExp(`^[\\.\\${path.sep}]+`), '');
+  const normalizedName =
+    normalizedUrl.length === 0
+      ? filename
+      : `${normalizedUrl.replace(pathSepPattern, '_')}_${filename}`;
+  const longname = `${normalizedName
     .toLowerCase()
     .replace(/[^a-z0-9_]/g, '')}.${type}`;
 


thymikee:
I've changed the base of PR to the next branch

jukben:
Nice catch!

thymikee:
Thanks @grahamlyus! We'll include this patch in next RC release.
Also, looks like we could cherry-pick this to `beta` channel, cc @satya164

grahamlyus:
@thymikee I can try adding a test when I get a chance. I would've done so but there are no existing tests for assetLoader

PR #438, opened 2018-05-29
Updating TypeScript configuration to the new style in Recipes

3vilguy:
Hello there!
Looks like there's a new configuration style for haul and unfortunately `Recipes.md` with TypeScript setup was forgotten. The old one still works but it's showing deprecation warning. I had a moment to play with it a bit more and eventually I got it to work. Doing this PR, in case anyone alse might find it useful.

### DIFF
--- a/docs/Recipes.md
+++ b/docs/Recipes.md
@@ -8,7 +8,7 @@ You will need to install `ts-loader` for Haul to work with TypeScript.
 
 ```yarn add --dev ts-loader```
 
-This is a `webpack.haul.js` that works with TypeScript.
+This is a `haul.config.js` that works with TypeScript.
 ```javascript
 import { createWebpackConfig } from "haul";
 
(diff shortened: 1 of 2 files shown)


docs/Recipes.md
```diff
@@ -10,31 +10,35 @@ You will need to install `ts-loader` for Haul to work with TypeScript.
 
 This is a `webpack.haul.js` that works with TypeScript.
```
zamotany: Could you also replace `webpack.haul.js` with `haul.config.js`?

3vilguy: Sure! It's done.
(On a side note, `yarn haul init` is still creating `webpack.haul.js`)

jukben: Thanks for the chatch @3vilguy I have created the issue for it #441

PR #440, opened 2018-05-30
Ss18/pr1 enhancement

ss18:
Fixed typos:
precident -> precedent
orginal -> original

Found with: https://github.com/ss18/grep-typos


jukben:
This will be in `next` as well I guess.

cc @thymikee

Issue #432, opened 2018-05-14
haul fails when installed via pnpm


tachtevrenidis:
Has anybody tried haul via pnpm instead of yarn or npm? I am having a hard time getting it to work...

pnpm: https://pnpm.js.org/en/

STEPS TO REPRO

* mkdir test-dir
* pnpm install haul --save-dev
* node node_modules/haul/bin/cli.js

### Current Behavior

Error:
```
node ./node_modules/haul/bin/cli.js 
module.js:538
    throw err;
    ^

Error: Cannot find module 'babel-polyfill'
    at Function.Module._resolveFilename (module.js:536:15)
    at Function.Module._load (module.js:466:25)
    at Module.require (module.js:579:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/Users/user/Development/pnpm-test/node_modules/.registry.npmjs.org/haul/1.0.0-rc.0/node_modules/haul/src/cli.js:10:1)
    at Module._compile (module.js:635:30)
    at Module._extensions..js (module.js:646:10)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/user/Development/pnpm-test/node_modules/.registry.npmjs.org/babel-register/6.26.0/node_modules/babel-register/lib/node.js:152:7)
    at Module.load (module.js:554:32)
    at tryModuleLoad (module.js:497:12)
```

### Expected Behavior
haul bundler starts running

### Haul Configuration (webpack.haul.js)
irrelevant for this issue

### Your Environment
| software         | version
| ---------------- | -------
| Haul             |  1.0.0-rc.0
| react-native     |  0.55.2
| node             |  8.9.3
| pnpmx      |  1.43.1

ThomasMichon:
I've done some investigations and made local workarounds in my product's codebase, which can hopefully translate here:

1. I updated `package.json` to include `babel-polyfill` in `dependencies`. Since `cli.js` is doing
```js
require('babel-polyfill');
```
directly, I think it is reasonable to declare it as a direct dependency.

2. I had to modify the regular expressions used for `babel-register` and `makeReactNativeConfig` to account for the fact that `node_modules/` might occur *multiple times* in a file's path.

```js
/node_modules\/(?!react|@expo|pretty-format|haul|metro)/
```
becomes
```js
/node_modules(?!.*[\/\\](react|@expo|pretty-format|haul|metro))/
```
and
```js
/node_modules(?!\/haul)/
```
becomes
```js
[...]

PR #443, opened 2018-06-19
pnpm support / fix debugging ui

chaseholland:
Added support for pnpm and fixing debugging ui. 

https://github.com/callstack/haul/issues/432

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "source-map": "^0.5.6",
     "strip-ansi": "^3.0.1",
     "thread-loader": "^1.1.5",
-    "webpack": "https://github.com/webpack/webpack.git#master",
+    "webpack": "^4.12.0",
     "webpack-hot-middleware": "^2.22.1",
     "ws": "^2.2.2",
     "xpipe": "^1.0.5"
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "babel-plugin-transform-flow-strip-types": "^6.22.0",
     "babel-plugin-transform-object-rest-spread": "^6.23.0",
     "babel-plugin-transform-react-jsx": "^6.24.1",
+    "babel-polyfill": "^6.26.0",
     "babel-preset-env": "^1.2.2",
     "babel-register": "^6.24.0",
     "camelcase-keys": "^4.0.0",
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "source-map": "^0.5.6",
     "strip-ansi": "^3.0.1",
     "thread-loader": "^1.1.5",
-    "webpack": "^4.6.0",
+    "webpack": "https://github.com/webpack/webpack.git#master",
     "webpack-hot-middleware": "^2.22.1",
     "ws": "^2.2.2",
     "xpipe": "^1.0.5"
--- a/src/babelRegister.js
+++ b/src/babelRegister.js
@@ -20,7 +20,7 @@ function resolve(presetsOrPlugins, type) {
  * To prevent `babel-register` from reading foreign Babel
  * configs, `babelrc` has to be set to `false` and Haul's
  * Babel config inlined into `babel-register`'s options.
- * 
+ *
  * Also presets and plugins need to be resolved to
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -37,6 +37,7 @@
     "babel-plugin-transform-flow-strip-types": "^6.22.0",
     "babel-plugin-transform-object-rest-spread": "^6.23.0",
     "babel-plugin-transform-react-jsx": "^6.24.1",
+    "babel-polyfill": "^6.26.0",
```
thymikee: whoops, good call on that

src/babelRegister.js
```diff
@@ -20,15 +20,17 @@ function resolve(presetsOrPlugins, type) {
...
  * Haul's `node_modules` directory, otherwise Babel would
  * search from them in project's `node_modules`.
  */
 require('babel-register')(
   Object.assign(
     {
-      ignore: /node_modules(?!\/haul)/,
+      // Replaces Haul's 'ignore' pattern, which does not work with PNPM.
+      // eslint-disable-next-line no-useless-escape
+      ignore: /node_modules(?!.*[\/\\]haul)/,
```
thymikee: Can you remove this useless escape inside character set? 
This will be enough
```
/node_modules(?!.*[/\\]haul)/
```

chaseholland: d'oh. Thanks!

src/babelRegister.js
```diff
@@ -20,15 +20,16 @@ function resolve(presetsOrPlugins, type) {
...
+ *
  * Also presets and plugins need to be resolved to
  * Haul's `node_modules` directory, otherwise Babel would
  * search from them in project's `node_modules`.
  */
 require('babel-register')(
   Object.assign(
     {
-      ignore: /node_modules(?!\/haul)/,
+      // Use a PNPM-compatible search pattern for node_modules.
```
ThomasMichon: It occurred to me that this could just be done:

```js
const haulSrcPath = __dirname; // Should be the 'haul/src' folder.
ignore: function (filename) {
    return filename.slice(0, haulPath) !== haulPath); // Ignore modules which are not Haul src files.
}
```

My understanding is that the point of the regex was to scope Babel to process only the Haul package contents, regardless of whether Haul is run by itself or as a dependency. Using the function callback pattern would save on Regex compute time!

src/utils/makeReactNativeConfig.js
```diff
@@ -92,7 +92,8 @@ const getDefaultConfig = ({
         { parser: { requireEnsure: false } },
         {
           test: /\.js$/,
-          exclude: /node_modules\/(?!react|@expo|pretty-format|haul|metro)/,
+          // eslint-disable-next-line no-useless-escape
+          exclude: /node_modules(?!.*[\/\\](react|@expo|pretty-format|haul|metro))/,
```
ThomasMichon: Similar to the `babel-register` `ignore` option, this can use a `function` instead:
```js
exclude: function (filename) {
    const normalizedPath = filename.replace('\\', '/');
    const indexOfNodeModules = normalizedPath.lastIndexOf('/node_modules/');

    for (const safeModule of ['react', '@expo', 'pretty-format', 'haul', 'metro']) {
        const indexOfSafeModule = normalizedPath.lastIndexOf(`/node_modules/${safeModule}/`);
        if (indexOfSafeModule > -1 && indexOfSafeModule >= indexOfNodeModules) {
            return false;
        }
    }

    if (indexOfNodeModules > -1) {
        return true;
    }

    return false;
}
```

That will probably out-perform the Regex, be more clear, and be more reliable.

thymikee: > That will probably out-perform the Regex

Not sure if we need perf gains in this area, regexps are pretty fast already, what's slow is Babel compilation

ThomasMichon: The risk right now is that the Regex will miss
`node_modules/random/haul23`
which is probably not desirable.

thymikee:
@ThomasMichon I'd be happy to accept a PR with the changes you suggested, especially when proven to be faster or easier to maintain/read :)

PR #448, opened 2018-07-10
feat: remove haul-integrate.sh, use CLI_PATH instead

### LINKED ISSUE #234: Deprecate `haul-integrate.sh`
There's CLI_PATH available: https://github.com/facebook/react-native/commit/dcecfa3189d7a71f57f908cc008e26c20a7cb2f5 for quite a few versions.

jukben:
Close #234 

This is just clean up of our codebase. Tested. Affects only "production" mode.


src/commands/init.js
```diff
@@ -322,79 +321,35 @@ const addToXcodeBuild = async (cwd: string) => {
...
-        'https://github.com/callstack-io/haul/blob/master/docs/Configuring%20Your%20Project.md#integrating-with-xcode'
-      )} for manual instructions`
+  if (countOccurrences(originalTask) !== 2) {
+    progress.warn(
+      `Couldn't edit Xcode project. Haven't recognized 'Bundle React Native code and images' build phase.`
     );
+    return;
   }
+
+  const haulTask = `shellScript = "# ${haulSignature}\nexport CLI_PATH=node_modules/haul/bin/cli.js\nexport NODE_BINARY=node`;
```
thymikee: can replace \n with visual newlines, as we're in template literal

src/commands/init.js
```diff
@@ -322,79 +321,35 @@ const addToXcodeBuild = async (cwd: string) => {
 
   let project = fs.readFileSync(path.join(entry, 'project.pbxproj')).toString();
 
-  const haulScriptKey = 'AD0CE2C91E925489006FC317';
+  const haulSignature = 'added by Haul';
+
+  /* Make sure we check both project iOS and iOS-TV, that's the magic behind "2" constant */
```
thymikee: how about making 2 a constant?
```
// Make sure we check both project iOS and iOS-TV, that's the magic behind "2" constant
const PROJECTS_COUNT = 2;
```

src/commands/init.js
jukben: let's leave it as one-liner I don't want to mess with Regex 👍

PR #451, opened 2018-07-11
Resolve path to `react-native` preset

jaridmargolin:
When running `haul` within a monorepo, the default behavior is to compile everything which is not in `node_modules` (with a few [exceptions](https://github.com/callstack/haul/blob/master/src/utils/makeReactNativeConfig.js#L78)). Webpack will use the realpath rather than the symlink path when processing rule `excludes`. As a result, any sibling packages within a monorepo gets compiled (this is desired behavior in my case).

Where the issue arrises, is that babel looks to resolve presets/plugins relative to the [file being transformed](https://github.com/babel/babel/issues/2974#issuecomment-155738271). It doesn't make sense for packages agnostic of distribution (web, native, etc...) to have to define `babel-preset-react-native` as a dependency. This should only be required within the react-native project itself.

By using an absolute path, we bypass any babel specific resolution logic, and instead resolve a single time relative to the project root using the native nodejs resolution algorithm.

### DIFF
--- a/src/utils/__tests__/__snapshots__/getBabelConfig.test.js.snap
+++ b/src/utils/__tests__/__snapshots__/getBabelConfig.test.js.snap
@@ -8,7 +8,7 @@ Object {
     "<<REPLACED>>/hot/babelPlugin.js",
   ],
   "presets": Array [
-    "react-native",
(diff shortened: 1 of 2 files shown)


thymikee:
I'm not opposed too, although didn't face such issues in a monorepo (not speaking about Haul though)

PR #452, opened 2018-07-13
Switch react-slate to scoped package

singingwolfboy:
I received these warnings when trying to install this package:

```
npm WARN deprecated react-slate-utils@0.2.0: Please use scoped package: @react-slate/utils
npm WARN deprecated react-slate@0.5.1: Please use scoped package: @react-slate/core
```

This pull request changes these dependencies to use the scoped packages, as requested.


thymikee:
@zamotany I think we're good with bumping the version, right?

singingwolfboy:
@thymikee I've updated the Yarn lockfile. The tests are failing, but I don't think those failures have anything to do with my changes. Do you know what's going on?

src/server/ui.js
```diff
@@ -10,8 +10,8 @@ const path = require('path');
 const React = require('react');
 const morgan = require('morgan');
 const { Subject } = require('rxjs');
-const { renderToTerminal, Text, ProgressBar } = require('react-slate');
-const { hideCursor, overwriteConsole } = require('react-slate-utils');
+const { renderToTerminal, Text, ProgressBar } = require('@react-slate/core');
```
zamotany: `ProgressBar` is a part of `@react-slate/components` package now, so you need to install it as well 😉

singingwolfboy: Good catch! Fixed.

thymikee:
Thanks you very much! :)

PR #453, opened 2018-07-13
Upgrade ws package

singingwolfboy:
This is a security fix: https://nodesecurity.io/advisories/550

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "thread-loader": "^1.1.5",
     "webpack": "^4.12.0",
     "webpack-hot-middleware": "^2.22.1",
-    "ws": "^2.2.2",
+    "ws": "^5.2.2",
     "xpipe": "^1.0.5"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -7674,7 +7674,7 @@ ws@^1.1.0:
     options ">=0.0.5"
     ultron "1.0.x"
 
-ws@^2.0.3, ws@^2.2.2:
+ws@^2.0.3:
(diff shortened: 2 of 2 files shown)


thymikee:
Can you also update the lockfile?

thymikee:
Also, can just migrate to v5 from v2? What about breaking changes?

singingwolfboy:
I've updated the lockfile now.

> Have you checked if we can just migrate to v5 from v2? What about breaking changes?

No, I admit that I haven't. I saw the security problem flagged by `npm audit`, and I made this pull request. It would probably be a good idea to have someone else take a look at this package -- someone more familiar with it.

singingwolfboy:
@jukben: I'm not very familiar with how to use the `ws` package, so I'm not the best candidate to do this upgrade. I don't want to accidentally break something! I just wanted to open this PR to alert the project maintainers that this dependency should be upgraded. If you know more about how this package works, and how `haul` uses it, maybe you should take another shot at doing this upgrade?

akrantz:
@jukben The security vulnerability is still present, and it would be good to resolve this. Could you please complete a pull request for https://github.com/callstack/haul/commits/jukben/ws-upgrade?

jukben:
@akrantz I'm sorry but currently I'm unable to finish it but I agree that it would be really good if someone starts where I left and smash it, finally.

tido64:
Sorry for reviving an old PR. We got alerted on some security issues as well and are basically in the same situation. Any chance we can get this going again?

I also see that `ws` has moved to 6.2.1 since.

cc @bweggersen @chaseholland

PR #454, opened 2018-07-13
Regenerate yarn.lock

singingwolfboy:
I wanted to update the `yarn.lock` file for #452 and #453, but I found I was unable to do so because I'm running Node 10:

```
$ yarn install
yarn install v1.7.0
(node:31541) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error upath@1.0.4: The engine "node" is incompatible with this module. Expected version ">=4 <=9".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

Regenerating this lockfile fetches a newer version of the `upath` dependency, which does not claim to be incompatible with Node 10.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,148 +2,230 @@
 # yarn lockfile v1
 
 
+"@babel/code-frame@7.0.0-beta.44":
+  version "7.0.0-beta.44"
+  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.0.0-beta.44.tgz#2a02643368de80916162be70865c97774f3adbd9"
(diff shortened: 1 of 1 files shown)


thymikee:
This is not a way to go. You should update just `upath` to latest, e.g. by running `yarn add upath`. This will update both package.json and yarn.lock properly

PR #455, opened 2018-07-13
Upgrade upath in lockfile

singingwolfboy:
Follow-up from #454. This resolves the following error on Node 10:

```
error upath@1.0.4: The engine "node" is incompatible with this module. Expected version ">=4 <=9".
```


thymikee:
Thank you!

PR #459, opened 2018-07-18
Fix minimization of bundles

### LINKED ISSUE #458: No minification in 1.0.0-rc.5 when bundling
### Current Behavior

no minification in 1.0.0-rc.5 when bundling:

```
react-native init Hello --version "0.53.0"
cd hello
yarn add -D haul
yarn haul init
yarn haul bundle --platform android --dev false --minify true --bundle-output dist/index.android.bundle --assets-dest dist 
head dist/index.android.bundle
/******/ (function(modules) { // webpackBootstrap
/******/  // The module cache
/******/  var installedModules = {};
/******/
/******/  // The require function
/******/  function __webpack_require__(moduleId) {
/******/
/******/    // Check if module is in cache
/******/    if(installedModules[moduleId]) {
/******/      return installedModules[moduleId].exports;
```

### Expected Behavior

bundle should be minified in production mode
### package.json ###
```{
  "name": "Hello",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "start": "node node_modules/react-native/local-cli/cli.js start",
    "test": "jest",
    "haul": "haul"
  },
  "dependencies": {
    "react": "16.2.0",
    "react-native": "0.53.0"
  },
  "devDependencies": {
    "babel-jest": "23.4.0",
    "babel-preset-react-native": "4.0.0",
    "haul": "^1.0.0-rc.5",
    "jest": "23.4.1",
    "react-test-renderer": "16.2.0"
  },
  "jest": {
    "preset": "react-native"
  }
}
```

```import { createWebpackConfig } from "haul";

export default {
  webpack: createWebpackConfig(({ platform }) => ({
    entry: `./index.js`
  }))
};
```
### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-rc.5
| react-native     |  0.53.0
| node             | 8.11.3
| npm or yarn      | yarn 1.7.0

appden:
The default `minimizer` created by Webpack does not provide a `test` options, so the default tests just for `.js` files. Since we rely on a different extension for our output bundles, we need to explicitly use `UglifyJsPlugin` ourselves.

Fixes #458

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "source-map": "^0.5.6",
     "strip-ansi": "^3.0.1",
     "thread-loader": "^1.1.5",
+    "uglifyjs-webpack-plugin": "^1.2.7",
     "webpack": "^4.12.0",
     "webpack-hot-middleware": "^2.22.1",
     "ws": "^2.2.2",
--- a/src/utils/__tests__/__snapshots__/getConfig.test.js.snap
+++ b/src/utils/__tests__/__snapshots__/getConfig.test.js.snap
@@ -5,7 +5,7 @@ exports[`creates config: diff ios/android config 1`] = `
 - First value
 + Second value
 
-@@ -44,25 +44,25 @@
+@@ -44,18 +44,18 @@
           \\"test\\": /\\\\.(aac|aiff|bmp|caf|gif|html|jpeg|jpg|m4a|m4v|mov|mp3|mp4|mpeg|mpg|obj|otf|pdf|png|psd|svg|ttf|wav|webm|webp)$/,
           \\"use\\": Object {
             \\"loader\\": \\"<<REPLACED>>/loaders/assetLoader.js\\",
@@ -24,6 +24,11 @@ exports[`creates config: diff ios/android config 1`] = `
     \\"optimization\\": Object {
       \\"concatenateModules\\": true,
       \\"minimize\\": false,
+      \\"minimizer\\": Array [
+        UglifyJsPlugin {
+@@ -80,11 +80,11 @@
+        },
+      ],
       \\"namedModules\\": true,
     },
     \\"output\\": Object {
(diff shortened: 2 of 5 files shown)


satya164:
Seems the plugin uses uglify-es which is unmaintained and causes bugs with React https://twitter.com/greweb/status/1013093979309727752?s=19

Is there a way to use V2? Maybe it'll be better to switch to babili?

hedgepigdaniel:
terser is a [drop in replacement](https://github.com/fabiosantoscode/terser#replacing-uglify-es-with-terser-in-a-project-using-yarn) for uglify (its a fork which is maintained). From what I understand, its faster than babili. Turning off compression makes the process about 3x faster again, with minimal/no increase in filesize.

appden:
Not sure which direction you guys would like to go here. Unfortunately, [babel-minify-webpack-plugin](https://github.com/webpack-contrib/babel-minify-webpack-plugin) doesn't look well maintained at this point either.

[Terser](https://github.com/fabiosantoscode/terser) looks great. We would either recommend users swap it in with the Yarn `resolutions` [hack](https://github.com/fabiosantoscode/terser#replacing-uglify-es-with-terser-in-a-project-using-yarn), or instead depend on this [terser-webpack-plugin](https://www.npmjs.com/package/@salesloft/terser-webpack-plugin) instead (**my personal recommendation**).

It looks like this will be temporary dependency since Webpack 5 will include its own `MinimizerWebpackPlugin` (according to [this comment](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/296#issuecomment-394827067)), which will support Terser.

drgx:
Hi @appden ,Thanks for you PR but i want to share my finding that related to haul minification:
- Latest haul version my bundle size was 7MB (minification not working)
- With you minify changes the bundle size reduce to 4MB
- Downgrade to haul 1.0.0-beta.13 (before webpack 4) the size reduce to 2.7MB
- Using metro bundle the bundle size 2.5MB
After your magnification changes, the bundle size still bigger than pervious haul version and metro.  😢 
Our android team really strict on the RN bundle size. Hopefully we can find a way to reduce the bundle size so it same or smaller than metro version 🙇

Great work btw! 👍

PR #461, opened 2018-07-24
Fixes crash when using live reload across multiple clients at the same time 

### LINKED ISSUE #415: Haul crashes when trying to (hot) reload 2 clients simultaneously
### Current Behavior
When I run 2 clients at the same time either with Live Reload or Hot Reload, Haul crashes with this stack trace:
```
/Users/thymikee/Projects/callstack/haul/src/server/index.js:58
    throw err;
    ^

TypeError: Cannot destructure property `item` of 'undefined' or 'null'.
    at TaskQueue.pop (/Users/thymikee/Projects/callstack/haul/src/compiler/TaskQueue.js:29:22)
    at Fork.fork.on (/Users/thymikee/Projects/callstack/haul/src/compiler/Compiler.js:98:54)
    at Fork.emit (events.js:127:13)
```

Which happens, because TaskQueue tries to pop an item that is not found there (likely removed a bit earlier by one of the clients).

### Expected Behavior
Haul should handle 2 clients gracefully

### Your Environment
Discovered while working on https://github.com/callstack/haul/pull/414

acoates-ms:
Fixes #415

### DIFF
--- a/src/compiler/Compiler.js
+++ b/src/compiler/Compiler.js
@@ -109,7 +109,7 @@ module.exports = class Compiler extends EventEmitter {
       // responses, so we put the task back with the same ID and decremented count.
       if (awaitingCount > 1) {
         this.tasks.set(taskId, { callback, awaitingCount: awaitingCount - 1 });
-      } else {
+      } else if (callback) {
         callback({
           errors: null,
           platform,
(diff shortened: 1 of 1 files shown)


thymikee:
@acoates-ms I've just published `1.0.0-rc.6`, let me know how it works for you :)

Issue #405, opened 2018-04-14
Remote debugging not working on android


hedgepigdaniel:
### Current Behavior

When starting an app with the setting "Debug JS Remotely" enabled, a RSOD appears with a java error which reads "unexpected url: /haul-hmr?platform=android"

I think the bug was introduced here: https://github.com/callstack/haul/pull/387/files

### Expected Behavior

Remote debugging should not stop the code from working

### Your Environment

React Native debugger 0.7.18
haul: https://github.com/hedgepigdaniel/haul/tree/b06ffb3bb84803ed08435eb3363a2468fee1da31 (based off recent master)
OS: Arch linux
OS: Android 8.1

From the console in react native debugger (the value of window.location does not seem to match the expectation in `polyfillEnvironment.js`.
```
window.location
WorkerLocation {href: "file:///opt/react-native-debugger/resources/app.asar/js/RNDebuggerWorker.js", origin: "file://", protocol: "file:", host: "", hostname: ""…}hash: ""host: ""hostname: ""href: "file:///opt/react-native-debugger/resources/app.asar/js/RNDebuggerWorker.js"origin: "file://"pathname: "/opt/react-native-debugger/resources/app.asar/js/RNDebuggerWorker.js"port: ""protocol: "file:"search: ""__proto__: WorkerLocation
```

`DEV_SERVER_ORIGIN` ends up being undefined.

This change fixes the issue: https://github.com/hedgepigdaniel/haul/commit/b4b464a6352b9219a10776e566f70c219d601063 but I'm not sure of the reason for using `window.location`

hedgepigdaniel:
Bump, this is still the case with 1.0.0-rc.5

stevefan1999:
https://github.com/callstack/haul/pull/464

hedgepigdaniel:
Fixed by #462

PR #462, opened 2018-07-29
Do not attempt to get the dev server origin from window.location

hedgepigdaniel:
See https://github.com/callstack/haul/issues/405

I'm not sure if this will break something else (or what the intention is when using `window.location`), but for me, this allows debugging to work with react-native-debugger and android with haul 1.0.0-rc.5.


zamotany:
AFAIK this will break remote debugging on Android

hedgepigdaniel:
Interesting - for me remote debugging on android is broken without this change.

Perhaps the issue is that I'm connecting to the debugger over a real network connection, rather than an ADB reversed port on localhost - I'll try with a USB cord and see what happens - perhaps there is a way to make it work in both cases.

hedgepigdaniel:
I've tested with both of these situations:
 * an ADB connection to the device over USB with `adb reverse tcp:8081 tcp:8081`, and the app configured with the "Dev server host and port for device" to `localhost:8081`
 * an ADB connection to the device over a network and the app configured with the "Dev server host and port for device" to `<ip of machine running haul + rn debugger>:8081`

In both of these cases, the app crashes when remote debugging is switched on as detailed in https://github.com/callstack/haul/issues/405. In both cases with this patch, it works correctly.

Other details I am testing with
 * React Native Debugger: 0.7.18
 * OS: Arch Linux
 * React Native: 0.54.4
 * Device: Nexus 6P Android 8.1.0

Not sure what else about the environment might vary such that it works/doesn't work the other way around?

hedgepigdaniel:
I should add that all these tests are done with HMR switched on - Since the crash happens when trying to connect the HMR socket, I think everything is fine in all cases unless you turn on HMR.

hedgepigdaniel:
OK how about this - it checks whether `window.location.protocol` is `file:` (It is, in my case), and only in that case it skips using `window.location` to set `DEV_SERVER_ORIGIN`. On other cases where `window.location` is the network location of the debug server, this should now not break debugging.

PR #464, opened 2018-07-30
Fix for RN Debugger

stevefan1999:
RN Debugger uses `file://` as its `window.location`.


stevefan1999:
Closed in favor of https://github.com/callstack/haul/pull/462

PR #465, opened 2018-07-31
Re-add missing class properties transform

stevefan1999:
There is a problem on this line:
https://github.com/callstack/haul/blob/d795af385dfb5e3e8a2ef5ceb405aff8ab15933a/src/hot/client/hotApi.js#L34-L36
Re-adding `transform-class-properties` resolved it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,6 +121,7 @@
     ],
     "plugins": [
       "transform-flow-strip-types",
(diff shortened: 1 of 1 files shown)


thymikee:
Can you also add it to package.json (and update yarn.lock)?

PR #466, opened 2018-08-01
Move artifact IO stream logs out of node_modules

stevefan1999-personal:
For some weird reasons, on Windows, these two artifact files, std{err|out}.log, are locked so I cannot proceed with my package installation while the server is online. It is better to leave it out of node_modules.

### DIFF
--- a/src/server/ui.js
+++ b/src/server/ui.js
@@ -17,8 +17,8 @@ const emoji = require('node-emoji');
 const Compiler = require('../compiler/Compiler');
 const logger = require('../logger');
 
-const outStream = 'node_modules/.haul-artifacts/stdout.log';
-const errStream = 'node_modules/.haul-artifacts/stderr.log';
+const outStream = '.haul-artifacts/stdout.log';
+const errStream = '.haul-artifacts/stderr.log';
 
 /**
  * Create and render React-powered UI for Haul packager server.


thymikee:
I'd rather investigate how to unlock them instead of moving out of node modules. 
With this change we'd need every user to git ignore these files, which is a pain.

stevefan1999:
@thymikee I'm afraid you can't. On Windows, if a file stream is created it will be locked to the handle owner process exclusively to prevent OS-level file buffer race condition unless you close the file for each line written, but it is extremely inefficient.

thymikee:
Let's add a config option `logsOutputPath` To Haul then :)

thymikee:
@stevefan1999-personal are you still interested in making this option happen? :)

thymikee:
No worries, there's no rush on that. Feel free to pick it up whenever you (or somebody else) feel comfortable. Marking as "good first issue"

stevefan1999-personal:
hmm, there's something i haven't noticed in 10 months...

zamotany:
@stevefan1999-personal yes, we've been working on a revamp of Haul and moving away from legacy implementation. In a new version, there are not artifacts in `node_modules`. Sorry that your PR wasn't merged and thank you for the time you've spent on this.

stevefan1999-personal:
@zamotany no worries🙂good luck on the overhaul

Issue #457, opened 2018-07-14
haul got lazy


ahfarmer:
### Current Behavior
Haul doesn't begin bundling when started with `--platform ios`.
<img width="298" alt="haul-lazy" src="https://user-images.githubusercontent.com/597825/42728395-3c2ff352-876e-11e8-9b09-23e17100df6f.png">

### Expected Behavior
Haul should begin bundling when started with `--platform ios`.

I think a feature was added to make it so that Haul doesn't begin bundling until a bundle is requested. But development flow would be faster if the iOS build and the bundling could happen simultaneously. If the platform is specified, haul knows what I want to build and could begin immediately.

Disabling this lazyloading would be particularly useful when working with a single platform (as I am).

zamotany:
In RC versions, there is no `--platform` option. Current behaviour is closes to how metro bundler works. However if you feel the need to support this, we strongly encourage you or anyone to contribute  by implementing this feature 😉

ahfarmer:
Cool - I'm also curious why haul doesn't just bundle everything when launched? Is there a concern about wasting CPU resources?

Would you accept a PR that changes haul to automatically bundle all platforms when launched?

appden:
I personally really like the laziness, since I'm mostly developing on iOS and rarely on Android, but I think it could be a command line option to disable that behavior (such as `--lazy false`). I don't think that would be a big code change.

tuncaulubilge:
A command line option to disable laziness would be great, but this will require platform to be defined, as the current structure bundles based on the platform that does the bundle request.

tuncaulubilge:
In the meanwhile, you can manually trigger bundling by requesting it via
`curl localhost:8081/index.android.bundle > /dev/null`

PR #467, opened 2018-08-01
Eager loading

stevefan1999:
This PR allows certain platform to start bundling prematurely

Fixes: https://github.com/callstack/haul/issues/457


jukben:
Thanks for the PR. What do you think about `eager` name? I think that it would suit better.

Also, it would be nice if it starts all platforms if you don't specify particular ones.

cc @zamotany

stevefan1999:
@jukben The problem is I don't know the set of valid platforms🤔But yeah I'd definitely call it `eager`.

jukben:
Wouldn't you mind to refresh the snapshots? (yarn jest --u)

stevefan1999:
ah, yes, the docs needs to be updated.

https://github.com/callstack/haul/blob/next/docs/CLI%20Commands.md

jukben:
We can handle it in next PR 👍

PR #477, opened 2018-09-17
Add support for Babel 7 + RN 0.57

hedgepigdaniel:
I've been testing this with React Native 0.57, Babel 7 stable, and Android. I'm not sure if it works with RN 0.56 and the corresponding Babel 7 beta, or on iOS.

## Changes
- Replace `babel-preset-react-native` with `metro-react-native-babel-preset`
- Upgrade all babel dependencies to equivalent v7 packages
- Update peerDependency on `react-native` to `^0.57`
- Slight refactor of `babelPlugin.js` to account for the changed API of `@babel/template`
- Changed the haste resolver plugin to account for the removal of `@providesModule` annotations in globally importable libs. Instead it includes all libraries in the `Libraries` directory and assumes that their file names are the name that should be imported. Not sure what is wrong with normal relative imports...
- Update all fixtures to use `react-native` 0.57.1
- Upgrade `react-hot-loader` to v4, and remove the additional monkey patching of `React.createElement` and co. HMR seems to work fine without it and the monkey patching no longer works on v4. For me it did not previously work with the extra monkey patching + `react-native-maps`, now it does, even though there is a scary error.
- Automatically add `react-hot-loader/patch` to entry points instead of patching it into the user's entry file with the babel plugin.


greg-dove:
fyi, I've been trying this along with using RN0.57.0 against our code which is a lerna based monorepo. So far, it has been a mostly good experience - thanks! 
I have however seen an issue that is new vs. the code immediately prior to these changes. It seems related to react-hot-loader. I was able to restore the previous behavior by making the following change in haul/hot/patch.js. I could be doing something 'quite wrong' here... I am just trying to address the issue (which was related to a comparison of scene.type in react-native-router-flux for Drawer). 

`var patch = require('react-hot-loader/patch');`
`if (__DEV__) patch.default.disableProxyCreation = true;`
`module.exports = patch;`

as opposed to your original:
`require('react-hot-loader/patch');`

I'm not suggesting this is a necessary change, because I am not at all familiar with this code - I'm just pointing out what I needed to do in order to restore my dev behavior for the codebase I am currently working on. If it's not correct, then it might be a helpful clue. Please let me know if I can provide any more information. 

(also, in terms of my testing: So far I have only been working with android on windows, I will try iOS tomorrow.)

chaseholland:
Expanding on what @greg-dove said above -- other than `react-hot-patch`, this is looking really good on iOS and Android for us. I was able to work around the hot patch issues by removing the entry file and plugin. Here are the relevant parts of my haul config in case it helps anyone else:

```js
        // Remove react-hot-patch as it causes issues with bundling
        config.entry = config.entry.filter((entry) => {
            return !entry.includes('hot/patch.js');
        });
```

```js
        // Remove the react-hot-loader config since we don't use it and it causes runtime issues
        config.module.rules.forEach((rule) => {
            if (Array.isArray(rule.use)) {
                rule.use.forEach((use) => {
                    if (use.options && use.options.plugins && Array.isArray(use.options.plugins)) {
                        use.options.plugins = use.options.plugins.filter((plugin) => { return !plugin.includes('react-hot-loader') });
                    }
                });
            }
        });
```

src/utils/makeReactNativeConfig.js
```diff
@@ -192,7 +192,7 @@ const getDefaultConfig = ({
         new HasteResolver({
           directories: providesModuleNodeModules
             ? providesModuleNodeModules.map(_ => moduleResolve(root, _))
-            : [moduleResolve(root, 'react-native')],
+            : [path.join(moduleResolve(root, 'react-native'), 'Libraries')],
```
acoates-ms: Something more is needed here.  
For users that specify a custom list of providesModuleNodeModules we need to ensure that at least react-native still gets scoped to the Libraries directory.

I wonder if providesModuleNodeModules option should be expanded to be able to include specific directories within the module.

Something like changing the type of providesModuleNodeModules to: ?
providesModuleNodeModules: (string | {module: string; directories: string[]})[]

acoates-ms: I changed the code locally to: 
```
        new HasteResolver({
          directories: (providesModuleNodeModules || [
            'react-native',
          ]).map(_ => {
            if (typeof _ === 'string') {
              if (_ === 'react-native') {
                return path.join(
                  moduleResolve(root, 'react-native'),
                  'Libraries'
                );
              }
              return moduleResolve(root, _);
            }
            return path.join(moduleResolve(root, _.module), _.directory);
          }),
          hasteOptions: hasteOptions || {},
        })
```
Which I think keeps the existing configs working, and allows additional modules to specify specific directories too.

hot/patch.js
```diff
@@ -8,6 +8,4 @@
 /**
  * Apply React patches in development by importing `haul/hot/patch`.
  */
-if (module.hot && process.env.NODE_ENV !== 'production') {
-  require('../src/hot/client/hotPatch');
-}
+require('react-hot-loader/patch');
```
zamotany: As @greg-dove said, we need to disable proxy creation, otherwise HMR won't persist the state:
```js
require('react-hot-loader/patch').default.disableProxyCreation = true;
```

jukben:
Great job everyone involved! ❤️

PR #481, opened 2018-10-10
Place the cache-loader cache inside the .cache subfolder within node_…

greg-dove:
Change the default cache location for the default 'cache-loader' to avoid it landing inside the local package and needing to add it to gitignore. 
Based on quick discussion with thymikee in chat room.


thymikee:
Thank you!

PR #483, opened 2018-10-13
Modify tests to work on windows

acoates-ms:
This makes a couple of minor modifications to allow the tests and snapshots to work on windows.

- Use `yarn.cmd` instead of `yarn` when using spawn on windows.
- Use `node haul/cli.js` instead of just `haul.cli.js` to run haul
- Use some additional substitutions when outputting the snapshot files, so that windows machines get the same result (Mostly logic around the root directory format, and of course the directory separators.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.js        text eol=lf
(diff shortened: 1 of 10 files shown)


greg-dove:
Just a quick check (not sure) - do the *.snap files also need to be added to gitattributes for windows eol=lf?

acoates-ms:
I didn't see issues without it.  My guess is that snap files are not seen as text files, and so don't automatically get converted to CRLF and stay LF.

PR #486, opened 2018-10-16
Ability to disable the prompt if the port is in use

voom-pairing-fenestron:
- This is useful for projects running haul within the context of
  a Procfile


voom-pairing-fenestron:
cc @hjhart @exmember

hjhart:
Sure! I'll work on a change and push it up.

Issue #487, opened 2018-10-18
Can't release application with latest haul


befront:
### Current Behavior
Unhandled JS Exception: undefined is not an object (evaluating 'r("./node_modules/react-hot-loader/patch.js").default.disableProxyCreation=!0')

### Expected Behavior
No errors

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-rc.8
| react-native     |  0.57.2
| node             | v9.11.1
| npm or yarn      | yarn 1.6.0

zamotany:
We are aware of this issue. The reason it is happening is that out `hot/patch.js` file is included in production build here:
https://github.com/callstack/haul/blob/next/src/utils/makeReactNativeConfig.js#L388

If anyone is interested in contributing, this file should be applied conditionally based on `dev` flag from here: https://github.com/callstack/haul/blob/next/src/utils/makeReactNativeConfig.js#L77

zamotany:
Should be fixed in #488

marchaos:
Still seeing this error on the latest version.

PR #488, opened 2018-10-18
bugfix: Can't release application with latest haul #487

greg-dove:
I will fix integration tests shortly.

### DIFF
--- a/src/utils/makeReactNativeConfig.js
+++ b/src/utils/makeReactNativeConfig.js
@@ -209,8 +209,7 @@ const getDefaultConfig = ({
         new HasteResolver({
           directories: (providesModuleNodeModules || [
             'react-native',
-          ]).map(
-            _ => {
+          ]).map(_ => {
             if (typeof _ === 'string') {
               if (_ === 'react-native') {
                 return path.join(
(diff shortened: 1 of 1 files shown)


src/utils/makeReactNativeConfig.js
```diff
@@ -372,22 +372,28 @@ function makeReactNativeConfig(
...
+  {
+    entry: userEntry,
+    root,
+    initializeCoreLocation = 'node_modules/react-native/Libraries/Core/InitializeCore.js',
+  }: {
+    entry: WebpackEntry,
+    root: string,
+    initializeCoreLocation?: string,
+  },
+  withHotPatch: boolean = true
```
thymikee: Instead of passing second argument, I'd pass the dev variable to the config object  and use that to determine whether add hot patch or not

greg-dove: Oh, got it. Will make that change...
I assume you mean like this:
```
entry: injectPolyfillIntoEntry(
        {
          root: options.root,
          initializeCoreLocation: options.initializeCoreLocation,
          entry,
	  dev: options.dev,
        }
      ),
```

Issue #490, opened 2018-10-23
`makeHot` needs to return same value with same params


bradennapier:
### Current Behavior

`makeHot`'s call result will ALWAYS return a different value.  This is a problem because it means that every little change will re-render the whole app when you want to enable hot reloading.  

In the case of `react-navigation` I was confused as-to why every time i toggled the drawer my entire screen was refreshing.  After many hours I finally realized it was due to `makeHot`.

This occurs because in the `makeHot` function the all of the function is creating a new React Component every time rather than returning the same one.  

See #491 for the PR to fix.

```javascript
const getHome = () => HomeView;

console.log(makeHot(getHome, 'Home')() === makeHot(getHome, 'Home')());
```

Since the above returns `false` it means that calling `makeHot` function is not returning the same value every time it is called - thus everything re-renders completely any time `getScreen` is called by `react-navigation`.

### Expected Behavior

```javascript
makeHot(() => HomeView, 'Home')() === makeHot(() => HomeView, 'Home')()
```

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-rc.9
| react-native     |  0.57.3
| node             |  10.12.0
| npm or yarn      | yarn

zamotany:
With `@haul-bundler/cli` we no longer support HMR due to various problems and issues, which we don't have a capacity to handle.

PR #491, opened 2018-10-23
fix makeHot

bradennapier:
This will fix the `makeHot` so it doesn't force-rerender every tiny little change that happens when used with routers.  It does this by not building a brand new component on every call of the resulting function. 

```javascript
const HotView = makeHot(() => HomeView, 'Home');
console.log(HotView() === HotView());
```

> Before the above would return false, therefore a re-render would be forced every time . `getScreen` is called by `react-navigation`

Note that this will still cause problems when utilized as documented since

```javascript
console.log(makeHot(() => HomeView, 'Home')() !== makeHot(() => HomeView, 'Home')())
```

Solution there would likely to use a `Map` of some sort to identify when we should return the previously created component instead of creating a new one.  Since I couldn't be 100% of the effects of doing this type of change I decided to at least make it fixable in the meantime. 

See #490

### DIFF
--- a/docs/HMR_react-navigation.md
+++ b/docs/HMR_react-navigation.md
@@ -73,7 +73,9 @@ const MyApp = StackNavigator({
 AppRegistry.registerComponent('MyApp', () => MyApp);
 ```
 
-Then, wrap those screen factories with `makeHot` call and pass the name of the screen as a second argument:
+Since a comparison is made of the results of `getScreen` each time the drawer is toggled, we need to make sure that subsequent calls of our function return the same value. 
+
+To do this, we need to create our `HotViews` then provide them as a result to `getScreen`.  
 
 ```diff
 // index.js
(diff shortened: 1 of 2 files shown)


bradennapier:
As a bit more - the way to make things work as needed with this setup would be to do the following:

```javascript
const HotViews = {
  Home: makeHot(() => HomeView, 'Home'),
};

const AppDrawerStack = createDrawerNavigator(
  {
    Home: {
      getScreen: HotViews.Home,
      navigationOptions: {
        drawerLabel: 'Home',
      },
    },
```

bradennapier:
Also updated docs to show the `HotViews` method since the previously described way would not work as desired without providing some sort of cache layer (good next step).

bradennapier:
Any plans for next release?

thymikee:
Released in `v1.0.0-rc.10`

PR #492, opened 2018-10-29
Update HMR_Setup.md

TNorth22:
Seems like a useful thing for us mobile native folks relatively new to RN.  Feel free to disagree :)


jukben:
It's already there https://github.com/callstack/haul/blob/next/package.json#L76

Issue #489, opened 2018-10-22
release build is not working


lavaxun:
### Current Behavior

I'm currently facing some issue with the latest haul & react native in RELEASE BUILD.
Please find the attached to understand the issue better.

I followed through the documentation and I managed to run the application in DEBUG BUILD.
You can reproduce the issue here: https://github.com/lavaxun/haul-issue.

HOWEVER it's not working well in the release build.
Screenshot below shows the build phase is using CLI_PATH. And it's giving hundreds of error when executing the bundling script during the build phase.
When I try to use 'fetch' to perform HTTP POST. it shows me 'unsupported bodyinit type' error.

<img width="1332" alt="not-working-with-cli_path" src="https://user-images.githubusercontent.com/1982412/47287994-b8a93b80-d626-11e8-8e7d-fba7736b3e36.png">

### Expected Behavior

Another screenshot below shows the bundling script is working fine in the build phase. but without specifying CLI_PATH.
And there is no problem in using 'fetch' to do HTTP POST.

<img width="1342" alt="working-without-cli_path" src="https://user-images.githubusercontent.com/1982412/47287991-b47d1e00-d626-11e8-835a-0536e3ed5572.png">

### Haul Configuration (haul.config.js)

```js
import { createWebpackConfig } from "haul";

export default {
  webpack: createWebpackConfig(({ platform }) => ({
    entry: `./index.js`
  }))
};
```

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-rc.9
| react-native     |  0.57.2
| react-native-navigation     |  1.1.491
| node             | 8.11.4
| npm or yarn      | 1.10.1

Summys:
+ 1

Summys:
+1

jeffreycgy:
Hi, I'm facing the same issue as well with the latest React Native version and Haul on Android. I'm able to run release build but I get this `unsupported BodyInit type` error while debug mode is working properly.

<img width="339" alt="screenshot 2018-10-29 at 9 40 00 am" src="https://user-images.githubusercontent.com/24579490/47625258-fd6d2f00-db5e-11e8-97c8-2e7a67c46951.png">

louiskhenghao:
+1

thymikee:
Would be great if somebody created a repo with a repro and investigate.

lavaxun:
@thymikee 
I have made a repo here https://github.com/lavaxun/haul-issue.
would be great if someone can help to investigate it.

[...]

PR #494, opened 2018-11-13
Fix missing Webpack banner in release mode

lavaxun:
This PR is a fix to the missing `fetch()` in release mode.
https://github.com/callstack/haul/issues/489

### DIFF
--- a/src/utils/makeReactNativeConfig.js
+++ b/src/utils/makeReactNativeConfig.js
@@ -160,6 +160,10 @@ const getDefaultConfig = ({
         minimize: !!minify,
         debug: dev,
       }),
+      new webpack.BannerPlugin({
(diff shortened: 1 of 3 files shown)


src/utils/makeReactNativeConfig.js
```diff
@@ -187,6 +187,10 @@ const getDefaultConfig = ({
               test: /\.(js|css|(js)?bundle)($|\?)/i,
               filename: '[file].map',
             }),
+            new webpack.BannerPlugin({
+              banner: 'if (this && !this.self) { this.self = this; };\n',
+              raw: true,
+            }),
```
thymikee: looks like this could be moved to common webpack config (along with the one on line 175)

lavaxun: Thanks for reminding.

PR #495, opened 2018-11-13
Do not transpile to commonjs to enable webpack tree shaking

Markionium:
This PR enables webpack's tree shaking by not allowing babel to transform `import`/`export` statements to commonjs modules.

https://webpack.js.org/guides/tree-shaking/

**More info**
We noticed we were getting bigger bundles when using metro vs haul. We use webpack and tree shaking for our web bundles which greatly reduces their size.

Since haul uses webpack and webpack takes care of the `import` and `export` statements I think it would make sense to let webpack take care of this and do it's magic?

We enabled this same setting on our mobile bundle by using the custom `.babelrc` option and saw a significant reduction in size of 366 KB from 2.17 MB to 1.80 MB.

There might be a reason why you haven't done this already, but I couldn't find any info around it in the current issues. I worked pretty smoothly for us so I figured I'd create a PR :)

### DIFF
--- a/src/utils/__tests__/__snapshots__/getBabelConfig.test.js.snap
+++ b/src/utils/__tests__/__snapshots__/getBabelConfig.test.js.snap
@@ -8,7 +8,12 @@ Object {
     "<<REPLACED>>/src/hot/babelPlugin.js",
   ],
   "presets": Array [
-    "<<NODE_MODULE>>/metro-react-native-babel-preset/src/index.js",
+    Array [
+      "<<NODE_MODULE>>/metro-react-native-babel-preset/src/index.js",
+      Object {
+        "disableImportExportTransform": true,
+      },
+    ],
   ],
 }
 `;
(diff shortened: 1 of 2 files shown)


jukben:
Thanks for that. Is there any downside of using this?

Markionium:
None that I'm currently aware of. The only difference is that it now removes unused code by doing static analysis on the files. Haul is able to do this because Webpack offers this feature. I am unfortunately not very familiar with Metro but a quick search on the repo seems to mention this is something they aim to add.
https://github.com/facebook/metro/issues/227
https://github.com/facebook/metro/pull/203

We tested this on our bundles for our iOS and Android app and didn't find any issues. If someone would encounter a bug with this then that would probably be a bug in Webpack's tree shaking and should then be reported there. Such a bug would likely also impact all web projects that use Webpack.

If someone would like to turn this off they could still use the `.babelrc` file to turn it off (like we now use it to enable this), but it seems like a sensible default to turn on.

tido64:
@zamotany, @thymikee: Is this something we can merge? We've seen a 300 kB reduction in bundle size with this change and tree shaking enabled.

thymikee:
Released as `v1.0.0-rc.13`

Issue #505, opened 2018-12-12
Unexpected keyword 'const'. Const declarations are not supported in strict mode.


liujb:
### Current Behavior

![](http://manhattan.didistatic.com/static/manhattan/do1_X5zkGJxiIk59kwkPLHzL)

### Expected Behavior

Expected not error.

### Haul Configuration (webpack.haul.js)

```
export default {
  webpack: createWebpackConfig(({ platform }) => ({
    entry: `./index.js`
  }))
};
```

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  v1.0.0-rc.10
| react-native     |  0.57.5
| node             |  v11.0.0
| npm or yarn      | 1.10.1

yongching:
I’m having this issue too, same environment.

hslavax:
I’m having this issue too, same environment. + 1

thymikee:
Can somebody provide a repro?

liujb:
@thymikee Sorry, could't provide repo now. When not use haul to start bundle server for android. App working correctly. 

I have setting haul.config.js like this, The problem is still.

```
import { createWebpackConfig } from "haul";
import path from 'path';

export default {
  webpack: createWebpackConfig(({ platform }) => ({
    entry: `./index.${platform}.js`,
    module: {
      rules: [
        {
          test: /\.js?$/,
          exclude: /(node_modules)/,
          include: [
            path.resolve(__dirname, './node_modules/react-navigation'),
            path.resolve(__dirname, './node_modules/query-string'),
            path.resolve(__dirname, './node_modules/strict-uri-encode'), 
            path.resolve(__dirname, './node_modules/decode-uri-component'),
            path.resolve(__dirname, './node_modules/react-navigation')
            path.resolve(__dirname, './node_modules/@react-navigation')
          ],
          use: {
            loader: 'babel-loader',
            options: {
              presets: ['@babel/preset-env'],
              plugins: ['@babel/plugin-proposal-object-rest-spread']
            }
          }
        }
      ]
    },
    resolve: {
      modules: ["node_modules"],
      alias: {
        'react-navigation': 'react-navigation/src/react-navigation.js'
      }
    }
  }))
};

```

liujb:
> @liujb [react-navigation@1.x](mailto:react-navigation@1.x) is too old, i'm using [react-navigation@3.x](mailto:react-navigation@3.x)

react-navigation@3.x have the problem by using haul.

thymikee:
[...]

Issue #400, opened 2018-04-06
Unexpected keyword 'const'. Const declarations are not supported in strict mode


naiyt:
### Current Behavior

When running a new app using haul with Typescript I get `Unexpected keyword 'const'. Const declarations are not supported in strict mode.`. This appears to have happened previously in https://github.com/callstack/haul/issues/159, but according to that ticket it was solved several versions ago by excluding `pretty-format` in 58c17ff. I'm on the latest version of haul and am seeing it happen.

### Expected Behavior

The error doesn't occur.

### Haul Configuration (webpack.haul.js)

I followed the [Typescript setup instructions](https://github.com/callstack/haul/blob/master/docs/Recipes.md):

`webpack.haul.js`:

```
module.exports = ({ platform }, { module, resolve }) => ({
  entry: `./index.js`,
  module: {
    ...module,
    rules: [
      {
        test: /\.tsx?$/,
        exclude: /node_modules/,
        use: [
          {
            loader: 'babel-loader',
          },
          {
            loader: 'ts-loader'
          },
        ],
      },
      ...module.rules
    ]
  },
  resolve: {
    ...resolve,
    extensions: [
      '.ts',
      '.tsx',
      `.${platform}.ts`,
      '.native.ts',
      `.${platform}.tsx`,
      '.native.tsx',
      ...resolve.extensions],
  },
});
```

`tsconfig.json`:

```
{
  "compilerOptions": {
    "declaration": false,
    "lib": ["es2015"],
    "module": "es6",
    "moduleResolution": "node",
    "jsx": "react",
    "target": "es2015",
    "allowSyntheticDefaultImports": true,
    "strictNullChecks": true,
    "noImplicitThis": true,
    "baseUrl": "."
  },
  "exclude": [
    "node_modules"
  ],
  "compileOnSave": false
}
```

`.babelrc`:

```
{
  "presets": ["react-native"]
}
```

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-beta.13
| react-native     |  0.55.1
| node             | both 9.11.1 and 8.11.1
| yarn      | 1.3.2

Thanks for any help!

deepsweet:
Happens to me as well since RN 0.54, and I found the problematic file: `node_modules/metro/src/lib/bundle-modules/MetroClient.js`, so now we have an additional lib to traspile explicitly.

Added `metro` to `node_modules/haul/src/utils/makeReactNativeConfig.js` and it works:

```js
[...]

PR #510, opened 2019-01-11
whitelist modules with const variables (to work out of the box with react-navigation)

alidcastano:
fixes https://github.com/callstack/haul/issues/505 (also related to https://github.com/callstack/haul/issues/502)

these were the modules I had to whitelist to get react-navigation working on a fresh react-native project. 

i also noticed #503 has not been released either, would be great to get these fixes in (or at least add it to recipes) since react-navigation is the recommended navigation library.

 in case anyone encounters this, here's a config you can use:

```
  const jsRule = config.module.rules.find(rule => rule.test && rule.test.test('foo.js'))
  jsRule.exclude = /node_modules(?!.*[\/\\](react|@react-navigation|query-string|strict-uri-encode|decode-uri-component|pretty-format|haul|metro))/
  
```

### DIFF
--- a/src/utils/makeReactNativeConfig.js
+++ b/src/utils/makeReactNativeConfig.js
@@ -97,7 +97,7 @@ const getDefaultConfig = ({
         {
           test: /\.js$/,
           // eslint-disable-next-line no-useless-escape
-          exclude: /node_modules(?!.*[\/\\](react|@react-navigation|@expo|pretty-format|haul|metro))/,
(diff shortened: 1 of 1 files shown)


thymikee:
Are these modules seriously not compiled to es5?

alidcastano:
looking at query-string source code, it contained `const` variables - not sure about others. also, only adding `react-navigation` worked for my ios simulator, but I had to add the other three to get it to work for my android one

@mortdatherobot reported something similar in https://github.com/callstack/haul/issues/400

Issue #493, opened 2018-11-06
Haul packager updates are very slow to load


dphaener:
### Current Behavior
When making a change to my source code and reloading, the app takes 5-10 seconds to actually download the changes and restart the app (this is after a 10 second webpack rebuild).

### Expected Behavior
The reloading should be just as fast as it is when using the metro bundler (nearly instantaneous)

### Haul Configuration (webpack.haul.js)
```javascript
import { createWebpackConfig } from 'haul';
import merge from 'webpack-merge';

const ALIASED_MODULES = [
  'react-native-linear-gradient',
  'galmorous-react-native',
  'react-native-svg',
  'lottie-react-native',
  'react-navigation',
];

export default {
  webpack: env => {
    const config = createWebpackConfig(({ platform }) => ({
      entry: ['babel-polyfill', `./index.${platform}.js`],
    }))(env);

    const moduleAliases = ALIASED_MODULES.reduce((acc, moduleName) => {
      acc[moduleName] = path.resolve(`./node_modules/${moduleName}`);
      return acc;
    }, {});

    const customConfig = {
      resolve: {
        symlinks: false,
        alias: moduleAliases,
      },
    };
    // eslint-disable-next-line no-useless-escape
    const customExclude = /node_modules(?!.*[\/\\](react|pretty-format|haul|metro|shared-mobile|glamorous-native|static-container))/;

    // Create a custom webpack configuration based on a deep merge of the default
    // configuration and our custom build settings from above.
    const jsLoader = config.module.rules.find(
      rule => rule.test && rule.test.test('foo.js')
    );

    jsLoader.exclude = customExclude;
    jsLoader.use.find(loader =>
      loader.loader.includes('thread')
    ).options.workerParallelJobs = 50;

    const mergedConfig = merge(config, customConfig);
    return mergedConfig;
  },
};
```

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-rc.9
| react-native     |  0.57.4
| node             | 10.0.0
| npm or yarn      | 6.4.1

tido64:
I had a look at this today since we're having the same issues. For our app, it takes between 6-12 seconds to reload, even when you manually trigger one.

As it turns out, what takes the most time is:
- A `stringify` call on the entire bundle here: [compiler/worker/initWorker.js:30](https://github.com/callstack/haul/blob/next/src/compiler/worker/initWorker.js#L30)
- Which gets sent over a WebSocket connection to the main process
[...]

PR #512, opened 2019-01-19
Drop memory-fs to speed up reloads.

tido64:
Since we need to serialize payloads to communicate with worker processes over
WebSockets, we will not see any improvements using an in-memory file system as
serializing big debug bundles is a time-consuming process in JS.

Before:
image

After:
image

Addresses #493.


tido64:
cc @thymikee, @zamotany

src/compiler/worker/initWorker.js
```diff
@@ -82,7 +89,7 @@ module.exports = function initWorker({
       if (fs.existsSync(filename)) {
         send(Events.FILE_RECEIVED, {
           taskId: payload.taskId,
-          file: fs.readFileSync(filename),
+          file: filename,
```
thymikee: I think it makes sense now to rename the field to `filename` or `filePath`, pick whatever feels better :)

src/compiler/worker/runWebpackCompiler.js
```diff
@@ -67,7 +67,6 @@ module.exports = function runWebpackCompiler({
     ],
   });
 
-  // Use memory fs
   compiler.outputFileSystem = fs;
```
thymikee: Am I missing something or it seems like we could remove the `outputFileSystem` entirely? https://webpack.js.org/api/node/#custom-file-systems

Or at least we could simplify https://github.com/callstack/haul/pull/512/files#diff-99e77f1c262ae763f3d7d593b2c0b417R44 by spreading fs and adding `mkdirp` and `join` if that's necessary

tido64: I added a comment here for future reference.

src/compiler/worker/runWebpackCompiler.js
```diff
@@ -67,6 +67,8 @@ module.exports = function runWebpackCompiler({
     ],
   });
 
+  // As of Webpack 4.12.0, |outputFileSystem| must be set since there is no
+  // fallback despite the documentation stating otherwise.
```
thymikee: Haha Webpack docs always being helpful on the topic 😄

tido64: Yeah, my first attempt was removing this line. I was surprised when it simply crashed 😄 

Thanks for reviewing and merging it 🎉

Issue #513, opened 2019-01-23
A lot ot "hot-update" files being generated


zaguiini:
### Current Behavior
Whenever I run the app, a lot of `hot-update.(json|js)` files are generated and they remain there after closing the app/bundler. **I am not even using hot reload. It's turned off on Dev Settings**

I can add them to `.gitignore` via `*.hot-update.*` and it's a "solution", but not ideal...

![image](https://user-images.githubusercontent.com/26530524/51603367-c35ac180-1ef0-11e9-9434-0d0beeb8dcac.png)

### Expected Behavior
No clutter. Or at least in another folder.

### Haul Configuration (webpack.haul.js)
```js
import {createWebpackConfig} from "haul";
import { resolve } from 'path'

export default {
  webpack: env => {
    const config = createWebpackConfig({
      entry: `./src/index.ts`,
    })(env);

    config.module.rules = [
      {
        test: /\.tsx?$/,
        exclude: '/node_modules/',
        use: [
          {
            loader: 'babel-loader',
          },
          {
            loader: 'ts-loader',
          }
        ]
      },
      ...config.module.rules,
    ];

    config.resolve.extensions = [
      '.ts',
      '.tsx',
      `.${env.platform}.ts`,
      '.native.ts',
      `.${env.platform}.tsx`,
      '.native.tsx',
      ...config.resolve.extensions,
    ]

    config.resolve.alias = {
      ...config.resolve.alias,
      src: resolve(__dirname, 'src'),
    }

    return config;
  }
};
```

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-rc.11
| react-native     |  0.57.8
| node             | v10.15.0
| npm or yarn      | 1.13.0

thymikee:
This is due to this change: https://github.com/callstack/haul/pull/512

We could move the updates to some temp directory (see: https://github.com/callstack/haul/pull/512/files#diff-99e77f1c262ae763f3d7d593b2c0b417R85), and provide a way to clean it up. 
Anybody willing to work on a solution?

apexskier:
Here's a proposed fix for this - https://github.com/callstack/haul/pull/516

sirajulm:
@apexskier  Thanks. This have fixed random folders being created. But the `documentation` for the `start` command is not updated with the new `--assetDest` option.

thymikee:
Mind sending a PR? :)

PR #516, opened 2019-02-11
feat: support assetsDest in start command

apexskier:
Resolves https://github.com/callstack/haul/issues/513

Adds an `--assets-dest` flag to `start` similar to the same flag for `bundle`. Defaults to a temporary directory on the filesystem, instead of the root directory of the project.

Tested both with and without the `--assets-dest` config flag.

### DIFF
--- a/src/__tests__/__snapshots__/cliEntry.test.js.snap
+++ b/src/__tests__/__snapshots__/cliEntry.test.js.snap
@@ -10,6 +10,7 @@ Options:
   --dev [true|false]               Whether to build in development mode, true by default
   --no-interactive [true|false]    Disables prompting the user if the port is already in use
   --minify [true|false]            Whether to minify the bundle, 'true' by default when dev=false
+  --assets-dest [string]           Path to directory where to store assets, eg. /tmp/dist.
   --config [string]                Path to config file, eg. haul.config.js, haul.config.js by
                                    default
   --eager [parse]                  Disable lazy building for platforms (list is separated by ','),
--- a/src/__tests__/__snapshots__/cliEntry.test.js.snap
+++ b/src/__tests__/__snapshots__/cliEntry.test.js.snap
@@ -10,7 +10,7 @@ Options:
   --dev [true|false]               Whether to build in development mode, true by default
   --no-interactive [true|false]    Disables prompting the user if the port is already in use
   --minify [true|false]            Whether to minify the bundle, 'true' by default when dev=false
-  --assets-dest [string]           Path to directory where to store assets, eg. /tmp/dist.
+  --assets-dest [string]           Path to directory where to store assets, eg. /tmp/dist
   --config [string]                Path to config file, eg. haul.config.js, haul.config.js by
                                    default
   --eager [parse]                  Disable lazy building for platforms (list is separated by ','),
--- a/src/__tests__/__snapshots__/cliEntry.test.js.snap
+++ b/src/__tests__/__snapshots__/cliEntry.test.js.snap
@@ -10,7 +10,7 @@ Options:
   --dev [true|false]               Whether to build in development mode, true by default
   --no-interactive [true|false]    Disables prompting the user if the port is already in use
   --minify [true|false]            Whether to minify the bundle, 'true' by default when dev=false
-  --assets-dest [string]           Path to directory where to store assets, eg. /tmp/dist
+  --assets-dest [string]           Path to directory where to store generated assets, eg. /tmp/dist
   --config [string]                Path to config file, eg. haul.config.js, haul.config.js by
                                    default
   --eager [parse]                  Disable lazy building for platforms (list is separated by ','),
--- a/src/commands/start.js
+++ b/src/commands/start.js
@@ -133,7 +133,7 @@ module.exports = ({
     },
     {
       name: 'assetsDest',
-      description: 'Path to directory where to store assets, eg. /tmp/dist.',
+      description: 'Path to directory where to store assets, eg. /tmp/dist',
     },
     {
       name: 'config',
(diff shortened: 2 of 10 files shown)


src/commands/start.js
```diff
@@ -130,6 +131,10 @@ module.exports = ({
         },
       ],
     },
+    {
+      name: 'assetsDest',
+      description: 'Path to directory where to store assets, eg. /tmp/dist.',
```
statm: super-nit: remove trailing period in description

src/__tests__/__snapshots__/cliEntry.test.js.snap
```diff
@@ -10,6 +10,7 @@ Options:
   --dev [true|false]               Whether to build in development mode, true by default
   --no-interactive [true|false]    Disables prompting the user if the port is already in use
   --minify [true|false]            Whether to minify the bundle, 'true' by default when dev=false
+  --assets-dest [string]           Path to directory where to store assets, eg. /tmp/dist.
```
statm: super-nit: remove trailing period in description

statm:
I'd suggest your alternative option (hide bundle artifacts via mktemp) as it preserves the original behavior instead of dumping tons of hot-update files under `root`, while allowing explicit override with `--assets-dest` if you do want to inspect them.

apexskier:
I hadn't considered behavior with respect to versioning. I think that makes it worthwhile to do the default.

src/compiler/worker/initWorker.js
```diff
@@ -34,6 +34,8 @@ module.exports = function initWorker({
     );
   }
 
+  const outputPath = JSON.parse(options).configOptions.assetsDest;
```
thymikee: it doesn't seem that you need to use `json.parse` here
```suggestion
  const outputPath = options.configOptions.assetsDest;
```

src/compiler/worker/runWebpackCompiler.js
```diff
@@ -50,6 +50,8 @@ module.exports = function runWebpackCompiler({
 
   let lastPercent = -1;
 
+  config.output.path = configOptions.assetsDest;
```
thymikee: I'd rather like to update the `getConfig` to handle `assetsDest` than mutating stuff like that

src/commands/start.js
```diff
@@ -130,6 +143,10 @@ module.exports = ({
         },
       ],
     },
+    {
+      name: 'assetsDest',
+      description: 'Path to directory where to store assets, eg. /tmp/dist',
```
thymikee:
```suggestion
      description: 'Path to directory where to store generated assets, eg. /tmp/dist',
```

src/commands/start.js
```diff
@@ -55,8 +58,18 @@ async function start(opts: *) {
   const directory = process.cwd();
   const configPath = getWebpackConfigPath(directory, opts.config);
 
+  let assetsDest;
+  if (opts.assetsDest) {
+    assetsDest = path.isAbsolute(opts.assetsDest)
+      ? opts.assetsDest
+      : path.join(directory, opts.assetsDest);
+  } else {
+    assetsDest = path.join(os.tmpdir(), fs.mkdtempSync('haul-start-'));
```
thymikee: I'm afraid `fs.mkdtempSync` will override the cache every time you run `haul start`, making Haul always start with a cold cache. Or am I getting it wrong?
Also you can move this assignment to line 61 and remove the `else` branch

statm: > I'm afraid `fs.mkdtempSync` will override the cache every time you run `haul start`, making Haul always start with a cold cache. Or am I getting it wrong?

Currently `haul start` does not re-use previously generated bundles - it always rebuilds the bundle (code cache) upon first request. IMO this is correct behavior as there might be source changes between 2 runs of `haul start`.

> Also you can move this assignment to line 61 and remove the `else` branch

Moving line 67 out of `else` would create a temp folder regardless of `opts.assetsDest`. That might not be ideal..

thymikee: You're right.

> IMO this is correct behavior as there might be source changes between 2 runs of `haul start`.

It should use cache whenever possible, but let's keep current behavior for now.

src/compiler/worker/initWorker.js
apexskier: From what I saw and experienced, this worker has its params injected as JSON, since it's a forked sub-process. They come from [here](https://github.com/callstack/haul/blob/next/src/compiler/createForkProcess.js#L38) and are also read [here](https://github.com/callstack/haul/blob/next/src/compiler/worker/runWebpackCompiler.js#L25).

thymikee: I misread this type definition, sorry!

src/utils/__tests__/__snapshots__/getConfig.test.js.snap
```diff
@@ -34,7 +34,7 @@ exports[`creates config: diff ios/android config 1`] = `
     \\"output\\": Object {
 -     \\"filename\\": \\"index.ios.bundle\\",
 +     \\"filename\\": \\"index.android.bundle\\",
-      \\"path\\": \\"<<REPLACED>>/src/utils/__tests__/fixtures\\",
+      \\"path\\": undefined,
```
thymikee: is this expected?

apexskier: Yes. This comes from https://github.com/callstack/haul/pull/516/files#diff-e573bff19b6673d9da63d58f54edd482R24 and is the reason this fix works.

apexskier: I actually don't want to pass this in tests, since they generate configs in a way I didn't anticipate.

thymikee:
Thanks for doing this @apexskier ❤️

Issue #519, opened 2019-02-17
Haul creating random folders in each run in project root


sirajulm:
### Current Behavior

Haul is pacakging the bundle and is being successfully loaded in both iOS and Android. There is no issue with the build, but haul is creating random folders in project `root`, prefixed as `haul-start-` while running Haul via `yarn haul start` command. 

Please refer the screenshot:
![image](https://user-images.githubusercontent.com/2260996/52916391-d2247080-3304-11e9-87bb-63e48f5e61c3.png)

 
### Expected Behavior

Haul should not be creating random folders each time it is being called, which would pollute the project root. If those folders are unavoidable either create single folder each time the command is run, which can then be added to `.gitignore` or provide option to the user to provide a path like `/dist` or `/build` to which these folders can be diverted.

### Haul Configuration (webpack.haul.js)

```
import { createWebpackConfig } from "haul";
import path from "path";

export default {
  webpack: (env, defaults) => {
    const config = createWebpackConfig({
      // ...defaults,
      entry: `./index.js`,
      resolve: {
        // ...defaults.resolve,
        alias: {
          // ...defaults.resolve.alias,
          react: path.join(__dirname, "node_modules/react"),
          "react-native": path.join(__dirname, "node_modules/react-native")
        }
      }
    })(env);

    config.module.rules = [
      {
        test: /\.tsx?$/,
        loader: "ts-loader"
      },
      ...config.module.rules
    ];

    config.resolve.extensions = [
      ".js",
      ".ts",
      ".tsx",
      `.${env.platform}.ts`,
      ".native.ts",
      `.${env.platform}.tsx`,
      ".native.tsx",
      ...config.resolve.extensions
    ];

    return config;
  }
};

```

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  ^1.0.0-rc.13
| react-native     |  0.58.4
| node             | v11.6.0
| npm or yarn      | 1.13.0

sirajulm:
After a lot of digging found the solution, needed to add `--assetDest` option to specify a destination. This is found missing the the documentation.

Related to Issue #513 which is fixed in PR #516 

Closing the issue.

thymikee:
Btw by default the asset destination is set to os tmp directory

sirajulm:
Yeah, have checked the code, and it points to tmp directory. But in Mac I'm getting the folders being created in the project root. Not sure why this happens. :-(

sirajulm:
[...]

PR #520, opened 2019-02-17
Fixing temp directory path for assetDest

sirajulm:
temp directory for assetDest was wrongly being created in project root, instead of `tmp`.

Fix for #519


thymikee:
@apexskier can you check that?

apexskier:
Ah, I must have messed up when testing this. Thanks for the fix! I've confirmed it works.

Issue #156, opened 2017-05-25
Document how to not exclude node_modules


alexanderkjeldaas:
### Current Behavior
The readme mentiones a few limitations, among them that files under node_modules are not transpiled.

### Expected Behavior
README.md includes either pointer or description of how to get `haul` to work, including how to ensure either node_modules are transpiled or how to configure webpack to do so.

junibrosas:
you can include the node_modules when you use a loader.

```
 include: [
    path.resolve(__dirname, "src"),
    path.resolve(__dirname, "node_modules/ui") // include specific package from node_modules
  ]
```

```
 include: [
    path.resolve(__dirname, "src"),
    path.resolve(__dirname, "node_modules/ui") // include the whole node_modules folder
  ]
```

Traviskn:
This is actually a bit more complex than it seems because the default haul config uses happypack (or the thread-loader in the most recent version) to transform js files in parallel with babel.  We'd need to remove and replace the existing js loader entirely.  If we don't want to lose out on the speed gain of happypack/thread-loader we'd also need to recreate that config.  It would be really helpful if we could somehow parameterize ignoring/including node_modules, it is (unfortunately) common for react native libraries to publish un-transformed code due to the default behavior of packager transforming node_modules

Traviskn:
Actually now that I'm looking at beta 7, it looks like there is no more exclude config?  Is this no longer ignoring node_modules?

Traviskn:
Newer releases are ignoring node modules now - This is a fairly hacky way to override the `exclude` config by reaching into the default rules and modifying them:

```javascript
// webpack.haul.js
module.exports = function getHaulWebpackConfig(options, defaults) {
  defaults.module.rules[1].exclude = undefined;

  return {
    ...defaults,
    entry: './index.js',
    module: {
      ...defaults.module,
      rules: [
        ...defaults.module.rules,
      ],
    },
  };
};
```

maxammann:
Currently having problems with using rn-fetch-blob. There is no easy solution to transpile specific modules currently.

chiubaka:
Any new information on this? The above discussion about internals that make it difficult to not exclude specific `node_modules` is concerning.

Currently having trouble with what I think is a transpilation issue with `aws-amplify-react-native`.

chiubaka:
Mostly I'm wondering if the above is still relevant, given that much has probably happened since Nov 10, 2017.

chiubaka:
[...]

PR #521, opened 2019-03-12
Add options to allow configuring of included node modules

chiubaka:
Seeking some early feedback on this PR. A little unsure about how/whether or not this change impacts `makeReactNativeConfig`, and whether or not it needs to. Also a little unsure of how the end user modifies the `EnvOptions` and whether or not this is really the best place to parameterize the new option. However, I think I've generally updated in the right place.

### Summary
Relates to #156.
* Updates the default Webpack configurations to resolve `.jsx` by default.
* Adds an env option to allow parameterized `node_module` includes to the existing JavaScript processing rule.
* **TODO:** Update Recipes to include an example of how to accomplish this, and link to this Recipe in the README

### Test plan
**TODO**: I'm thinking I will update `src/utils/__tests__/makeReactNativeConfig.test.js` to ensure that passing the new options to `createWebpackConfig` results in a passing diff against a newly created snapshot.


src/utils/makeReactNativeConfig.js
```diff
@@ -31,6 +31,7 @@ type ConfigOptions = {|
   providesModuleNodeModules?: (string | { name: string, directory: string })[],
   hasteOptions?: *,
   initializeCoreLocation?: string,
+  includedNodeModules?: Array<string>,
```
tido64: Shouldn't this be `excludedNodeModules`?

src/utils/makeReactNativeConfig.js
```diff
@@ -69,6 +70,30 @@ type DEPRECATEDWebpackConfigFactory = (
...
+];
+
+/**
+ * Constructs a RegExp for the default JavaScript processing rule
+ */
+const getJavaScriptExcludeRegex = (
+  includedNodeModules?: Array<string>
+): RegExp => {
+  includedNodeModules = includedNodeModules || [];
+  includedNodeModules = includedNodeModules.concat(
```
tido64: I'd use a spread operator here to make things a bit neater:
```js
  const modules = [
    ...(includedNodeModules || []),
    ...DEFAULT_INCLUDED_NODE_MODULES,
  ];
```

chiubaka:
Sorry for the slow response--just returning from vacation.

src/utils/makeReactNativeConfig.js
chiubaka: The way things have been implemented before this PR is not the typical "include all node modules except the ones listed for exclusion", but rather a "include none of the node modules except the ones listed for exclusion."

I haven't changed the paradigm, I've just exposed these as a configurable option.

src/utils/makeReactNativeConfig.js
```diff
@@ -52,6 +52,29 @@ type DEPRECATEDWebpackConfigFactory = (
...
+  '@expo',
+  'pretty-format',
+  'haul',
+  'metro',
+];
+
+/**
+ * Constructs a RegExp for the default JavaScript processing rule
+ */
+const getJavaScriptExcludeRegex = (
```
tido64: Nit: I would rename this to something else. It's not strictly creating an exclude regex though it might be used as such.

src/utils/__tests__/__snapshots__/getConfig.test.js.snap
```diff
@@ -81,7 +81,7 @@ Object {
       },
       Object {
         "exclude": /node_modules\\(\\?!\\.\\*\\[\\\\/\\\\\\\\\\]\\(react\\|@react-navigation\\|@react-native-community\\|@expo\\|pretty-format\\|haul\\|metro\\)\\)/,
```
tido64: Could you also add tests and diff against this snapshot?

src/utils/makeReactNativeConfig.js
```diff
@@ -67,6 +90,7 @@ const getDefaultConfig = (options: EnvOptions): WebpackConfig => {
     providesModuleNodeModules,
     hasteOptions,
     disableHotReloading,
+    includedNodeModules,
```
tido64: Sorry, I'm still a Webpack noob. It seems that we're preventing these modules from being transpiled by babel-loader. My concern is that it's not obvious what this option does without having to dig into the code. Can we rename it to something more specific, e.g. `babelLoaderExcludedModules`?

@thymikee, what do you think?

thymikee: I think I would be more keen on exposing the helper and document how to use it instead of introducing a new option

chiubaka: Just for my understanding as an outsider: would you mind elaborating on your thinking about the trade-offs here?

chiubaka:
Hey all, just wanted to bump this!

PR #522, opened 2019-03-19
Added flag for disabling hot reloading.

tido64:
Allows hot reloading to disabled with a flag `--hot-reloading` set to `false`.

### Summary

In our environment, hot reloading has never worked reliably and may in some cases cause crashes. We've gone so far as to have scripts to fix and strip away rules, modules, etc. from the generated configuration. Adding a flag to Haul to disable hot reloading is the right solution™.

### Test plan

- Start Haul: `yarn haul start -- --hot-reloading false`
- Make sure that the dev server still works as expected with Hot Reloading turned off

### DIFF
--- a/src/cliEntry.js
+++ b/src/cliEntry.js
@@ -7,7 +7,6 @@
 import type { Command } from './types';
 
 const minimist = require('minimist');
-const camelcaseKeys = require('camelcase-keys');
 const clear = require('clear');
 const inquirer = require('inquirer');
 const path = require('path');
@@ -160,11 +159,9 @@ async function run(args: Array<string>) {
 
   const opts = command.options || [];
 
-  const { _, ...flags } = camelcaseKeys(
-    minimist(args, {
-      string: opts.map(opt => opt.name),
-    })
-  );
+  const { _, ...flags } = minimist(args, {
+    string: opts.map(opt => opt.name),
+  });
 
   if (command.adjustOptions) command.adjustOptions(flags);
 
--- a/src/cliEntry.js
+++ b/src/cliEntry.js
@@ -7,6 +7,7 @@
 import type { Command } from './types';
 
 const minimist = require('minimist');
+const camelcaseKeys = require('camelcase-keys');
 const clear = require('clear');
 const inquirer = require('inquirer');
 const path = require('path');
@@ -159,9 +160,11 @@ async function run(args: Array<string>) {
 
   const opts = command.options || [];
 
-  const { _, ...flags } = minimist(args, {
-    string: opts.map(opt => opt.name),
-  });
+  const { _, ...flags } = camelcaseKeys(
+    minimist(args, {
+      string: opts.map(opt => opt.name),
+    })
+  );
 
   if (command.adjustOptions) command.adjustOptions(flags);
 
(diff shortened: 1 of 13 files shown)


tido64:
cc @acoates-ms @thymikee

src/cliEntry.js
```diff
@@ -160,11 +159,9 @@ async function run(args: Array<string>) {
 
   const opts = command.options || [];
 
-  const { _, ...flags } = camelcaseKeys(
```
thymikee: Why removing this?

src/utils/getBabelConfig.js
```diff
@@ -18,7 +18,7 @@ const DEFAULT_BABELRC = {
   ],
 };
 
-module.exports = function getBabelConfig(cwd: string) {
+module.exports = function getBabelConfig(cwd: string, options: Object) {
```
thymikee: Please make the type strict

src/utils/__tests__/__snapshots__/getConfig.test.js.snap
```diff
@@ -242,3 +242,182 @@ Object {
   "target": "webworker",
 }
 `;
+
+exports[`creates config: ios config with hot reloading disabled 1`] = `
```
thymikee: This snapshot is too big. Please use snapshot-diff and compare default config to this

src/cliEntry.js
tido64: `camelcaseKeys` renamed `hot-reloading` to `hotReloading`. This later fails in `validateOptions()` because we're trying to parse a flag that doesn't exist ([see line 97](https://github.com/callstack/haul/blob/ee773d160861d2831bf28512322a680fa970668a/src/cliEntry.js#L97)).

tido64: Aha, I see where this is used now… Let me try and fix `validateOptions()` instead.

tido64: Okay, I think I just misunderstood how to name the options. It's fixed now. Thanks for catching this :)

tido64:
@thymikee Care to take another look? 😄

src/utils/makeReactNativeConfig.js
```diff
@@ -381,19 +376,21 @@ function injectPolyfillIntoEntry({
...
 }) {
   const reactNativeHaulEntries = [
     ...getPolyfills(),
     require.resolve(path.join(root, initializeCoreLocation)),
-    require.resolve('./polyfillEnvironment.js'),
   ];
 
-  if (dev) {
+  if (dev && !disableHotReloading) {
+    reactNativeHaulEntries.push(require.resolve('./polyfillEnvironment.js'));
```
thymikee: Why polyfills are only applied in dev now?

tido64: Correct me if I’m wrong, but it seems like it only sets up the environment for hot reloading. I’m happy to move it back otherwise.

thymikee: Looks like we could require this from hot patch to avoid confusion

tido64: Good idea. I think we should also rename it while we’re at it. I could do this in a separate PR if you don’t mind. Don’t want to scope creep on this one. 😀

thymikee: Let's do this

tido64: Thanks! Follow up here: #527

Issue #514, opened 2019-01-25
Xcode shell script errors after update to Babel 7 / RN 0.57


nbennink:
### Current Behavior
I'm upgrading from React Native 0.56 to React Native 0.57. I have to update babel 7 to make this work. So I followed [these](https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md#updating-to-this-version) steps from the React Native changelog. And some of the tips in #411. But I get a lot of errors from the haul shell script when I build in xCode afterwards.

![Xcode](https://i.imgur.com/Q5sZNCl.png)
(and +/- 1200 more like this)

I'm using version `0.45.1` of `metro-react-native-babel-preset` in my `.babelrc` as suggested by the React Native upgrade guide. But the same issue appears when using `babel-preset-react-native` (both V4 and V5).

### Expected Behavior
There is no exit code so the build completes "successful" and everything in the app seems to work. But preferably without all the errors in xcode. I still have issue #487 when I make a release build which should be fixed with this haul version.

### Haul Configuration (webpack.haul.js)
```
let srcAppPath = path.join(__dirname, 'src');
let nodeModulesPath = path.join(__dirname, 'node_modules');

module.exports = ({platform}, defaults) => {
    const config = {
        entry: `./index.js`,
        resolve: {
            ...defaults.resolve,
            plugins: [...defaults.resolve.plugins],
            modules: [srcAppPath, nodeModulesPath],
            extensions: ['.native.js', '.js', '.ejs', '.jsx', '.less', 'html', 'json', `.${platform}.js`, ...defaults.resolve.extensions],
            alias: {
                _COMPONENTS: 'js/components',
                _LIBS: 'js',
                _CONFIG: 'config',
            },
        },
    };

    return config;
};

```

### Your Environment
| software         | version
| ---------------- | -------
| Haul             |  RC-09, RC-10 and RC-11 (tried multiple versions)
| react-native     |  0.57.8
| node             |  10.15.0
| npm or yarn      | npm 6.4.1

nbennink:
Thanks for your response. That doesn't work unfortunately. The error seems to be called from this script: Bundle React Native code and images
```
# added by Haul
export CLI_PATH=node_modules/haul/bin/cli.js
export NODE_BINARY=node
../node_modules/react-native/scripts/react-native-xcode.sh
```

And not from the haul session that is running at that moment.

tido64:
[...]

PR #525, opened 2019-03-21
Avoid tripping Xcode's script error detection.

tido64:
Xcode seems to fail if the verbosity level is set to anything other than `verbose`. Some of the characters used in the other levels makes Xcode think that errors have occurred. This was filed as #514 but was closed a bit early since there is a workaround.

### Summary

The patch detects whether we're in interactive mode and applies the correct verbosity level.

### Test plan

By default, without adding `--progress verbose` to the bundle command, Xcode should not fail.

### DIFF
--- a/src/commands/bundle.js
+++ b/src/commands/bundle.js
@@ -214,7 +214,11 @@ let command = ({
       name: 'progress',
       description:
         'Display bundle compilation progress with different verbosity levels',
-      default: 'compact',
+      default: () => {
+        // Ensure that we don't trip Xcode's error detection. 'verbose' is the
+        // only level that doesn't make Xcode think that the bundle failed.
+        return !process.stdin.isTTY ? 'verbose' : 'compact';
+      },
       example: 'haul bundle --progress minimal',
       choices: [
         {


tido64:
@nbennink, @tkow: Will you be able to test whether this works for you without setting `EXTRA_PACKAGER_ARGS`?

cc @thymikee

src/commands/bundle.js
```diff
@@ -214,7 +214,11 @@ let command = ({
       name: 'progress',
       description:
         'Display bundle compilation progress with different verbosity levels',
-      default: 'compact',
+      default: () => {
+        // Ensure that we don't trip Xcode's error detection. 'verbose' is the
+        // only level that doesn't make Xcode think that the bundle failed.
+        return !process.stdin.isTTY ? 'verbose' : 'compact';
```
hjhart: I propose defaulting to `expanded` instead of `verbose`. On my set up, `expanded` and `verbose` had the exact same amount of lines (2846), but I feel like expanded conveys the opposite of `compact`.

Also, not sure if the comments are necessary – If someone reviews the PR and commit message that should be sufficient for understanding the reasons behind the change.

tido64: I tried `expanded` first, and it caused issues when run in Azure DevOps. The comment is clearer as lines may be moved by later commits and will be harder to track down.

PR #527, opened 2019-03-25
Moved polyfillEnvironment.js to avoid confusion.

tido64:
This is a follow up to https://github.com/callstack/haul/pull/522#discussion_r268521093.

### Summary

`polyfillEnvironment.js` is only used to provide extra information for HMR to correctly resolve server origin. Its name and location currently implies otherwise. To minimize confusion, it is moved under `hot/` and required only in `hot/patch.js`.

### Test plan

HMR should work as it used to.


tido64:
cc @thymikee

hot/patch.js
```diff
@@ -1,10 +1,12 @@
 /**
  * Copyright 2017-present, Callstack.
  * All rights reserved.
- * 
+ *
  * @flow
  */
 
+require('../src/hot/client/polyfillEnvironment');
```
thymikee: actually, why not inline all of this? it's really not that much code, but requiring through `../src` creates weird relations and will be an issue (possibly hard to catch without manually testing) when we decide to build the project instead of doing awful babel-register

tido64: I hit these two issues when I tried to inline them. Moving the files into the same folder works though ¯\\_(ツ)_/¯

```
$ yarn test
yarn run v1.15.2
$ yarn flow && yarn lint && yarn jest
$ flow --include-warnings
Error: hot/patch.js:8
  8: import { NativeModules } from 'react-native';
                                   ^^^^^^^^^^^^^^ react-native. Required module not found

Error: hot/patch.js:61
 61:     __webpack_require__.p = `${global.DEV_SERVER_ORIGIN}/`; // eslint-disable-line no-undef
         ^^^^^^^^^^^^^^^^^^^ __webpack_require__. Could not resolve name


Found 2 errors
```

tido64:
Hi, @zamotany. Let me know how I can improve this PR. Thanks 😄

PR #528, opened 2019-03-25
Quiet warnings from Babel regarding support for corejs v3

afoxman-microsoft:
### Summary

Haul uses @babel/preset-env without specifying a corejs version. By default, babel uses corejs v2. With babel 7.4, there is now the option to use corejs v3.

Babel intends to transition to corejs v3 by default in the future. To get developers ready, they are emitting a warning that the default will change, and that a specific corejs version should be specified.

This change makes the use of corejs v2 explicit, silencing the warnings. The Haul development team can move to corejs v3 in the future, but that is a much larger project.

[Babel announcement](https://babeljs.io/blog/2019/03/19/7.4.0)

### Test plan

This change should not alter the build in any way. I verified that builds continue to succeed and produce the correct artifacts, without emitting the warning from babel.


thymikee:
Looks like we need to update to latest Babel to have this option present? Can you take care of it?

afoxman-microsoft:
Yes, definitely. I just caught that too. I'll update to >= 7.4 and update the PR.

PR #530, opened 2019-04-05
Disable namedModules in prod to reduce bundle size.

tido64:
`namedModules` tells webpack to use readable module identifiers for better debugging.
See https://webpack.js.org/configuration/optimization/#optimizationnamedmodules.

### Summary

We should disable it in production mode to reduce the bundle size.

### Test plan

The output bundle should work as before.


tido64:
cc @thymikee @acoates-ms

thymikee:
Thank you!

tido64:
@thymikee Thanks for the quick response ❤️

Issue #498, opened 2018-11-22
Calls to `logger.info` in assets loader causes the webpack logger to restart


bweggersen:
### Current Behavior
When I run `haul bundle` I see over 100 lines with the text `Webpack: Starting ...`. The interruption of the webpack logger is caused by calls to `logger.info` inside `loaders/assetLoader.js#L153`.

Sample log:

```
 INFO  loaded babel-loader configuration: [my-path]/.babelrc

 INFO  Assets Destination: [my-path]

 INFO  Bundle output: index.mobile.bundle

 INFO  Bundle output (resolved): [my-path]/index.mobile.bundle
Webpack: Starting ...

 INFO  Haul is now bundling your React Native app in production mode.

Webpack: Starting ...

Webpack: Starting ...

Webpack: Starting ... (repeated over 100 times)

...

  ✔ Compile modules
  ✔ Build modules
  ✔ Optimize modules
  ✔ Emit files

Webpack: Finished after 14.548 seconds.
```

### Expected Behavior
When I run `haul bundle` the logger should not restart multiple times.

Sample log:

```
 INFO  loaded babel-loader configuration: [my-path]/.babelrc

 INFO  Assets Destination: [my-path]

 INFO  Bundle output: index.mobile.bundle

 INFO  Bundle output (resolved): [my-path]/index.mobile.bundle

 INFO  Haul is now bundling your React Native app in production mode.

  ✔ Compile modules
  ✔ Build modules
  ✔ Optimize modules
  ✔ Emit files

Webpack: Finished after 14.548 seconds.
```

### Your Environment

| software         | version
| ---------------- | -------
| Haul             |  1.0.0-rc.10
| react-native     |  0.57.0
| node             |  8.12.0
| npm or yarn      |  yarn 1.12.1

PR #533, opened 2019-04-15
Mark asset path logs conditional on --verbose

bweggersen:
### Summary
When the asset loader evaluates the sizes of a new asset, it will also log the path to the asset. Whenever the logger outputs a new log line during a Webpack build, Webpack will repeat the following log line: `Webpack: Starting ...`. This PR adds makes the logging of the asset path conditional on the `--verbose` flag, similar to other log events in the same file.

Before:
```
Webpack: Starting ...

Webpack: Starting ...

Webpack: Starting ...

Webpack: Starting ...

Webpack: Starting ...

Webpack: Starting ...

Webpack: Starting ...

Webpack: Starting ...

Webpack: Starting ...

Webpack: Starting ...

Webpack: Starting ...

  √ Compile modules
  > Build modules (96%)
    → 1271 of 1322 modules :: [path]
```

After
```
Webpack: Starting ...

  √ Compile modules
  √ Compile modules
  √ Build modules
  √ Optimize modules
  √ Emit files
```

### Test plan
This change will only decrease the amount of log lines.

### DIFF
--- a/src/loaders/assetLoader.js
+++ b/src/loaders/assetLoader.js
@@ -195,7 +195,9 @@ module.exports = async function assetLoader() {
   };
 
   try {
-    logger.info(`path: ${this.resourcePath}`);
+    if (isVerbose) {
+      logger.info(`path: ${this.resourcePath}`);
+    }
 
     info = size(this.resourcePath);
 


bweggersen:
Ref.: #498

PR #541, opened 2019-05-06
feat: dynamic import in development bundle

zamotany:


packages/haul-core/src/webpack/plugins/ReactNativeEnvPlugin.ts
```diff
@@ -0,0 +1,34 @@
...
+          return `${asyncEval}\n${source}`;
+        }
+      );
+
+      (compilation.mainTemplate as any).requireEnsure.tap(
+        'ReactNativeEnvPlugin',
+        (source: string) => {
+          return source.replace(
+            /importScripts\((.+)\)/gm,
+            'return asyncEval(__webpack_require__.p + $1)'
```
thymikee: is `return` necessary here?

zamotany: yes, it is extremely necessary

thymikee: a comment on why would be helpful, or a test showing before & after, especially that you replace all occurrences of `importScripts` (`g` flag in regex). I assume it's always at the end of some function block, but would be cool to make it more obvious :)

PR #546, opened 2019-05-07
[FEAT]inline sourcemap option

dratwas:


packages/haul-preset-0.59/src/defaultConfig.ts
```diff
@@ -15,7 +15,8 @@ import getBabelConfigPath from './getBabelConfigPath';
 
 export default function getDefaultConfig(
   runtime: Runtime,
-  options: EnvOptions
+  options: EnvOptions,
+  inlineSourceMap?: boolean
```
zamotany: How about we just pass a whole `haulConfig` here?

dratwas: Good idea :+1:

PR #548, opened 2019-05-08
feat: source map generation for (non-minified) RAM bundle

zamotany:
https://github.com/callstack/haul/projects/2#card-21106029


packages/haul-ram-bundle-webpack-plugin/package.json
```diff
@@ -14,17 +14,24 @@
...
+    "webpack-sources": "^1.3.0"
   },
   "dependencies": {
     "@haul/core": "1.0.0-rc.15",
     "mkdirp": "^0.5.1"
   },
   "devDependencies": {
     "@types/mkdirp": "^0.5.2",
+    "@types/webpack-sources": "^0.1.5",
+    "@babel/core": "7.4.4",
```
thymikee: is this version locked on purpose? creates duplicated entries in lockfile

PR #549, opened 2019-05-09
fix: use correct global when evaluating modules in RAM bundle on Android

zamotany:
https://github.com/callstack/haul/projects/2#card-21226745


zamotany:
Closing this PR since it is not invalid anymore.

PR #551, opened 2019-05-09
docs: update docs for @haul/*

zamotany:
https://github.com/callstack/haul/projects/2#card-21119516


docs/Configuration.md
```diff
@@ -10,7 +10,7 @@ title: Configuration
 The `haul.config.js` file is a configuration used by `Haul`. The simplest config looks like this:
 
 ```js
-import { createWebpackConfig } from "haul";
+import { createWebpackConfig } from "@haul/preset-0.59";
```
thymikee: is the idea to have a preset per RN version?

packages/haul-ram-bundle-webpack-plugin/README.md
```diff
@@ -0,0 +1,58 @@
...
+[![PRs Welcome][prs-welcome-badge]][prs-welcome]
+[![MIT License][license-badge]][license]
+[![Chat][chat-badge]][chat]
+[![Code of Conduct][coc-badge]][coc]
+
+RAM bundle plugin for Haul and Webpack. This module is intended for internal use.
+
+__When using `@haul/cli ram-bundle` command, this plugin will be added and configured automatically__, based on passed options.
+
+You can read moe about Haul here: https://github.com/callstack/haul.
```
thymikee:
```suggestion
You can read more about Haul here: https://github.com/callstack/haul.
```

docs/Configuration.md
zamotany: Generally yes. In case where `preset-0.59` would work just fine for `0.60.x` we can create a table with preset to RN version compatibility

zamotany: If you have any other idea how to handle version differences between RN, I'm open for suggestions.

thymikee: I think this is fine. A bit more maintenance needed, but the differences are usually gonna be negligible so it shouldn't be that bad. One thing I could advice is to store presets in a directory so the repo is cleaner when RN hits 70 or 80 😄: `@haul/presets/0.59`

PR #555, opened 2019-05-13
feat: add file ram bundle support

zamotany:
https://github.com/callstack/haul/projects/2#card-21105966

### DIFF
--- a/packages/haul-ram-bundle-webpack-plugin/src/FileRamBundle.ts
+++ b/packages/haul-ram-bundle-webpack-plugin/src/FileRamBundle.ts
@@ -6,19 +6,11 @@ import { Module } from './WebpackRamBundlePlugin';
 import { countLines } from './utils';
 
 export default class FileRamBundle {
-  bootstrap: string;
-  modules: Module[];
-  sourceMap: boolean;
-
   constructor(
-    bootstrap: string,
-    modules: Module[],
-    sourceMap: boolean = false
-  ) {
-    this.bootstrap = bootstrap;
-    this.modules = modules;
-    this.sourceMap = sourceMap;
-  }
+    public bootstrap: string,
+    public modules: Module[],
+    public sourceMap: boolean = false
+  ) {}
 
   build({
     outputDest,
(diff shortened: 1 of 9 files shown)


packages/haul-ram-bundle-webpack-plugin/src/FileRamBundle.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  constructor(
+    bootstrap: string,
+    modules: Module[],
+    sourceMap: boolean = false
+  ) {
+    this.bootstrap = bootstrap;
+    this.modules = modules;
+    this.sourceMap = sourceMap;
+  }
```
dratwas: WDYT about ?
```js
constructor(
    public bootstrap: string,
    public modules: Module[],
    public sourceMap: boolean = false
  )
```

packages/haul-ram-bundle-webpack-plugin/src/__tests__/FileRamBundle.test.ts
```diff
@@ -0,0 +1,43 @@
+import { RawSource } from 'webpack-sources';
+import MAGIC_NUMBER from 'metro/src/shared/output/RamBundle/magic-number';
+import FileRamBundle from '../FileRamBundle';
+
+test('FileRamBundle should create valid RAM bundle', () => {
+  const bootstrapper = '(function() { /* bootstrap */ })()';
+  const modules = [
+    {
+      id: 0,
```
thymikee: likely not relevant to the test, but looks like `id` can be a string, maybe worth adding a case where it actually is a string

zamotany: it's tested in integration tests

packages/haul-ram-bundle-webpack-plugin/src/__tests__/integration.test.ts
```diff
@@ -213,30 +247,154 @@ describe('should build and successfully evaluate bundle', () => {
...
+
+    await wait(1);
+
+    expect(global.crash).toBeDefined();
+    expect(() => {
+      ((global as any) as { crash: Function }).crash();
+    }).toThrow('test error');
+  });
+
+  it('with namedModules:true', async () => {
```
thymikee: there's quite some duplicated code here. consider using `test.each` for less repetition

PR #557, opened 2019-05-13
[FEAT] Add inspectors flags to command arguments

dratwas:
Support `--node-inspector` and `--haul-inspector` args


zamotany:
@thymikee We plan to revisit docs soon

PR #565, opened 2019-05-29
fix: transpile async/await in `haul-bundler/babel-preset-react-native`

zamotany:
Native async/await behave incorrectly on RN 0.59 on both platforms and they not always resolve or resolve under certain conditions.


thymikee:
This is interesting. Have you encountered a bug filed in webkit for that maybe?

PR #570, opened 2019-06-11
feat: config revamp

zamotany:


packages/haul-cli/src/commands/shared/prepareWebpackConfig.ts
```diff
@@ -1,10 +1,10 @@
-import path from 'path';
+// import path from 'path';
```
dratwas: Shouldn't it be removed?

packages/haul-ram-bundle-webpack-plugin/src/WebpackRamBundlePlugin.ts
```diff
@@ -290,52 +277,52 @@ export default class WebpackRamBundlePlugin {
...
 
-  generateDebugFiles(
-    moduleMappings: ModuleMappings,
-    bootstrapCode: string,
-    extraData: Object
-  ) {
-    if (!this.config.debug) {
-      return;
-    }
+  // generateDebugFiles(
```
dratwas: Will it be removed or what?

PR #573, opened 2019-06-17
feature: multi bundle support in packager server

zamotany:


thymikee:
Tests are still failing

zamotany:
@thymikee I know, and it's because of some weird thing happening on CI, cause they are passing locally and we are using the `start` command (for which the integration test if failing) and we can clearly see it's working. Because of that we decided to temporarily skip this failing test and fix it later, since we are running short on time and have other higher-priority tasks to do.

PR #575, opened 2019-06-18
fix: add member arrow function with `super` transform

zamotany:


packages/haul-babel-preset-react-native/src/transforms/superMemberArrowFunction.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export default function superMemberArrowFunction() {
+  return {
+    visitor: {
+      Program(programPath: babel.NodePath<babel.types.Program>) {
+        programPath.traverse({
+          Super(path) {
+            let parentFunctionPath = path.getFunctionParent();
+
+            // Check if parent function is a async arrow function as a class property.
```
dratwas: Should be `an async` but TBH are you really mean `async`?

zamotany: `an async` needs to be removed

thymikee:
Is this some kind of babel-loader bug? Maybe there's a tracking issue somewhere?

PR #584, opened 2019-07-02
Optimisation babel transform + fixes

zamotany:
* Babel transform to strip unused `Platform.select` platform code when creating static bundle
* Remove unnecessary `__webpack_require__.p` from async chunk loading code
* Fix response from debugger worker when communicating through bridge

### DIFF
--- a/packages/haul-babel-preset-react-native/src/index.ts
+++ b/packages/haul-babel-preset-react-native/src/index.ts
@@ -30,7 +30,16 @@ export default function getHaulBabelPreset() {
         plugins: [require('@babel/plugin-transform-flow-strip-types')],
       },
       {
-        plugins: defaultPlugins,
+        plugins: defaultPlugins.concat(
+          process.env.HAUL_PLATFORM
+            ? [
+                [
+                  require('./transforms/stripDeadPlatformSelect'),
+                  { platform: process.env.HAUL_PLATFORM },
+                ],
+              ]
+            : []
+        ),
       },
       {
         test: /node_modules\/react-native/,
--- a/packages/haul-babel-preset-react-native/src/transforms/__tests__/stripDeadPlatformSelect.test.ts
+++ b/packages/haul-babel-preset-react-native/src/transforms/__tests__/stripDeadPlatformSelect.test.ts
@@ -1,0 +1,134 @@
+import stripDeadPlatformSelect from '../stripDeadPlatformSelect';
+import { transformSync } from '@babel/core';
+
+const fixture1 = `
+const value = Platform.select({
+  android: 'android',
+  ios: () => {},
+  default: { test: true },
+  web: 'what?'
+});
+`;
+
+const fixture2 = `
+const defaults = { default: 'default' };
+const value = Platform.select({
+  ios: () => {},
+  ...defaults,
+});
+`;
+
+const fixture3 = `
+const value = Platform.select({
+  android: () => {},
+});
+`;
+
+const fixture4 = `
+const value = Platform.select({
+  ios: false,
+  ['andr' + 'oid']: () => {},
+});
+`;
+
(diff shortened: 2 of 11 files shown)


packages/haul-babel-preset-react-native/src/transforms/__tests__/stripDeadPlatformSelect.test.ts
```diff
@@ -0,0 +1,134 @@
...
+  it('should replace Platform.select with value for default case if there is no matching platform', () => {
+    const { code } = transformSync(fixture1, {
+      plugins: [[stripDeadPlatformSelect, { platform: 'custom' }]],
+      babelrc: false,
+      configFile: false,
+    }) || { code: '' };
+
+    expect(code).toMatch(/const value = {\s*test: true\s*}/);
+    expect(code).not.toMatch('Platform.select');
+    expect(code).toMatchSnapshot();
```
thymikee: I'd use `toMatchInlineSnapshot` but whatever

packages/haul-babel-preset-react-native/src/transforms/__tests__/stripDeadPlatformSelect.test.ts
```diff
@@ -0,0 +1,134 @@
...
+const fixture6 = `
+const value = Platform.select({
+  ios: false,
+  android() {}
+});
+`;
+
+describe('stripDeadPlatformSelect transform', () => {
+  it('should replace Platform.select with value for default case if there is no matching platform', () => {
+    const { code } = transformSync(fixture1, {
```
thymikee: also grouping fixtures is harder for readability, can you move them inside the test cases?

packages/haul-babel-preset-react-native/src/index.ts
```diff
@@ -30,7 +30,16 @@ export default function getHaulBabelPreset() {
         plugins: [require('@babel/plugin-transform-flow-strip-types')],
       },
       {
-        plugins: defaultPlugins,
+        plugins: defaultPlugins.concat(
+          process.env.HAUL_PLATFORM
```
thymikee: Can you make `platform` a preset config? This is hacky

packages/haul-core/assets/debuggerWorker.js
```diff
@@ -117,16 +117,18 @@ class DebuggerWorker {
...
     } catch (e) {
+      error = e;
       if (self.ErrorUtils) {
         self.ErrorUtils.reportFatalError(e);
       } else {
         console.error(e);
       }
     } finally {
-      sendReply(null, null);
+      sendReply(null, error);
```
thymikee: I guess it doesn't matter if error is `null` or `undefined`?

packages/haul-babel-preset-react-native/src/index.ts
zamotany: It is but I cannot pass this value (`platform`) in a normal way as a preset option.

packages/haul-core/assets/debuggerWorker.js
zamotany: It does, it has to have some value or be `undefined`. `null` is incorrectly handled on iOS.

packages/haul-babel-preset-react-native/src/index.ts
thymikee: wtf Babel

zamotany: I'm not sure it's babel fault, `platform` is like a dynamic value, we cannot hardcode it in the Babel config nor we can pass additional metadata to Babel plugins through `babel-loader`.

thymikee: can you document it somehow somewhere?

PR #587, opened 2019-07-09
feat: add platform specific configuration options

zamotany:


packages/haul-core/src/preset/makeConfigFactory.ts
```diff
@@ -143,125 +152,57 @@ export default function makeConfigFactory(getDefaultConfig: GetDefaultConfig) {
...
-            );
-          });
+        if (env.bundleMode === 'single-bundle') {
+          applySingleBundleTweaks(
+            env,
+            normalizedTemplatesConfig,
+            normalizedBundleConfig,
+            webpackConfig
+          );
+        } else {
```
thymikee:
```suggestion
        } else if (env.bundleMode === 'multi-bundle') P
```

packages/haul-core/src/preset/utils/getSourceMapPlugin.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  if (bundleConfig.sourceMap === 'inline') {
+    return new webpack.EvalSourceMapDevToolPlugin({
+      ...baseOptions,
+      publicPath: `http://${serverConfig.host}:${serverConfig.port}/`,
+    } as webpack.EvalSourceMapDevToolPluginOptions);
+  } else if (bundleConfig.sourceMap) {
+    return new webpack.SourceMapDevToolPlugin({
+      ...baseOptions,
+      publicPath: `http://${serverConfig.host}:${serverConfig.port}/`,
```
thymikee: publicPath is the same, is this expected? if so, you could fit it into baseOptions

PR #588, opened 2019-07-11
feat: replace ink with terminal-kit and revamp server UI and DX

zamotany:
- switch from `ink`/`reac`t to `terminal-kit` which is more robust
- fix compilation progress reporting
- log errors to main screen buffer in terminal where server crashes (previously it was logging it to alternate screen buffer, which disappeared after switching back to main screen buffer)


thymikee:
Do you have some screenshots, gifs?

zamotany:
@thymikee 
Kapture 2019-07-11 at 14 55 07

thymikee:
Moving away from React feels like a step backwards. What were the issues with it? How about using react-slate instead of ink?

zamotany:
@thymikee 
- `react-slate` is out-of-date and doesn't properly handle mouse scrolling
- `ink` has weird bugs with layout, and can cut off the logs
- `ink` can crash for some unknown reason (I don't remember in which case unfortunately)
- neither `react-slate` nor `ink` are stable and robust

thymikee:
Too bad!

PR #591, opened 2019-07-15
docs: revamp for new config

zamotany:


docs/CLI Commands.md
```diff
@@ -33,22 +29,30 @@ yarn haul start --eager # Prebuild bundle for all platforms
...
 
 Whether to minify the bundle while serving. This will make the bundling process slow due to extra minification step, and is `false` by default.
 
-### `--port [number]`
+#### `--port [number]`
 
 Port on which the server should run. defaults to `8081`.
 
-### `--config [path]`
+#### `--assets-dest <string>`
```
joeblynch: I notice `--assets-dest` is described below as being where assets are stored for the build output, is this intended to be named something related to temp files instead?

docs/CLI Commands.md
```diff
@@ -57,26 +61,82 @@ Example: `haul bundle --platform android --bundle-output build/index.bundle --as
...
+
+Path to the directory to store the assets. e.g. - `build/assets`.
+
+#### `--sourcemap-output <string>`
+
+Path to use for a source map file, e.g. - `build/index.android.bundle.map`.
+
+#### `--config [path]`
+
+Path to the webpack haul config. defaults to `haul.config.js`.
```
thymikee:
```suggestion
Path to the webpack haul config. Defaults to `haul.config.js`.
```

thymikee: Also, is it necessary to mention webpack here?

docs/Recipes.md
```diff
@@ -56,101 +67,98 @@ And a corresponding (example) `tsconfig.json`
 }
 ```
 
+Please note, that if you want to use synthetic defaults imports, aka things like `import React from 'react'` instead of `import * as React from 'react'` you will need to:
```
thymikee:
```suggestion
Please note, that if you want to use synthetic default imports, a.k.a. things like `import React from 'react'` instead of `import * as React from 'react'` you will need to:
```

docs/Configuration.md
```diff
@@ -3,70 +3,375 @@ id: configuration
...
+  },
+});
 ```
 
-> The `createWebpackConfig` helper function provided by Haul takes an object of settings – currently just `entry`, or function which takes an object (type Object) specified below as the first argument – it should return an object of settings as well. 
+This config will allow you to run `bundle`, `ram-bundle` and `multi-bundle` (**experimental**) with `--platform custom`.
+
+Remember that, if the `--bundle-output` is provided, the `templates.filename.custom` won't be used, instead the filename will be inferred from the `--bundle-output` argument.
+
+Please note that by providing `platforms` array, no defaults will be added, meaning `ios` and `android` are no loger supported. To add support for `ios` and `android` alongside `custom` use `['custom', 'ios', 'android']`:
```
thymikee:
```suggestion
> Please note that by providing `platforms` array, no defaults will be added, meaning `ios` and `android` are no loger supported. To add support for `ios` and `android` alongside `custom` use `['custom', 'ios', 'android']`:
```

docs/Configuration.md
```diff
@@ -3,70 +3,375 @@ id: configuration
...
+- `app0` bundle with application logic (as Indexed RAM bundle).
+- `app1` bundle with application logic (as Indexed RAM bundle).
 
-    config.plugins.push(new CaseSensitivePathsPlugin());
+To build static bundles run `multi-bundle` command instead of `bundle` or `ram-bundle` command. When running `start` command the packager server will automatically detect the multi-bundle mode, so no actions needs to be done here.
 
-    return config;
+The host logic is up to the user to provide, based on the use case, but here's a example implementation:
+
+```js
```
thymikee:
```suggestion
```jsx
```

PR #595, opened 2019-07-17
Fix /open-stack-frame validation

koke:
### Summary

React Native is sending a methodName in the payload that fails validation.

### Test plan

To reproduce, throw an exception somewhere in your React Native app. From the error screen, tap in one of the stack frames. Haul's log before this would show `POST /open-stack-frame 400` and do nothing.

After this change, it should open your editor in the file referenced in the stack frame.


zamotany:
@koke Thank you for your PR 👏

zamotany:
I'll do a release today.

PR #598, opened 2019-07-17
feat: add missing reload command

zamotany:


packages/haul-cli/src/commands/reload.ts
```diff
@@ -11,10 +13,40 @@ export default function reloadCommand(runtime: Runtime) {
...
+          const req = http.request(requestOptions, () => {
+            runtime.logger.done('Sent reload request');
+            resolve();
+          });
+
+          req.on('error', e => {
+            const error = e.toString();
+            if (error.includes('connect ECONNREFUSED')) {
+              runtime.logger.error(
+                `Reload request failed. Make sure Haul is up.`
```
thymikee: This message is not cool. What does it mean that "Haul is up"? When can you get the connection refuse error?

PR #599, opened 2019-07-17
feat: add --output-file and --json options to allow logging to a file

zamotany:


packages/haul-core/src/runtime/Logger.ts
```diff
@@ -98,6 +104,18 @@ export default class Logger {
         this.inspectorClient.emitEvent(new LoggerEvent(level, args));
       }
 
+      if (this.logFile !== undefined) {
+        fs.appendFileSync(
+          this.logFile,
+          (this.logAsJson
+            ? JSON.stringify({ timestamp: new Date(), level, messages: args })
```
thymikee: why not iso here?

zamotany: `JSON.stringify` will convert it autmatically:
```
JSON.stringify(new Date())
""2019-07-17T15:57:39.215Z""
```

thymikee: TIL

PR #609, opened 2019-07-22
fix: --port option not being used (fixes #607)

### LINKED ISSUE #607: `haul start --port` not being used
## Environment

```
$ react-native --version
react-native-cli: 2.0.1
react-native: 0.60.4
$ yarn haul --version
yarn run v1.17.3
$ haul start --version
0.9.0
✨  Done in 1.16s.
```

## Description

Command: `yarn haul start --port 3030` (https://github.com/callstack/haul/blob/64802a6174b4949c999397673af58ed1ad59ebe1/docs/CLI%20Commands.md#haul-start)

Expected behavior: `haul` should start on port 3030 and we should be able to open http://localhost:3030/.

Actual behavior: `haul` started on the default port 8081.

Screen Shot 2019-07-19 at 12 18 10 PM

Similarly, `yarn haul start --no-interactive true` also doesn't work. It still displays the UI.

## Reproducible Demo

`next` branch

```
react-native init MyProject
cd MyProject
yarn add --dev @haul-bundler/cli
yarn haul init
yarn haul start --port 3030
```
I have also tried `yarn haul start --port=3030` and that doesn't work too.

zamotany:
fixes #607

### DIFF
--- a/packages/haul-core/src/server/UserInterface.ts
+++ b/packages/haul-core/src/server/UserInterface.ts
@@ -99,7 +99,7 @@ export default class UserInterface {
 
     platforms.forEach((platform, index) => {
       this.compilations[platform] = {
-        width: Math.min(110, this.terminal.width) - (platformNameLength + 22),
+        width: Math.min(110, this.terminal.width) - (platformNameLength + 20),
         value: 0,
         title: `${' '.repeat(
           platformNameLength - platform.length
--- a/packages/haul-core/src/server/UserInterface.ts
+++ b/packages/haul-core/src/server/UserInterface.ts
@@ -99,7 +99,7 @@ export default class UserInterface {
 
     platforms.forEach((platform, index) => {
       this.compilations[platform] = {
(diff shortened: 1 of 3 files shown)


packages/haul-core/src/server/UserInterface.ts
```diff
@@ -99,7 +99,7 @@ export default class UserInterface {
 
     platforms.forEach((platform, index) => {
       this.compilations[platform] = {
-        width: Math.min(110, this.terminal.width) - (platformNameLength + 20),
+        width: Math.min(110, this.terminal.width) - (platformNameLength + 22),
```
thymikee: this should go in a separate commit, but I guess you're aware :P

zamotany: Well ideally this should go into separate PR

zamotany:
```suggestion
        width: Math.min(110, this.terminal.width) - (platformNameLength + 20),
```

PR #610, opened 2019-07-22
fix: replace --tempDir with `--temp-dir`

zamotany:


thymikee:
Any documentation that needs updating?

zamotany:
@thymikee nope, the docs already use `--temp-dir`

PR #611, opened 2019-07-22
feat: better handle interactivity in start command

zamotany:


thymikee:
FYI yargs supports negating params with `--no-`: https://github.com/yargs/yargs/blob/master/docs/tricks.md#negate
So having `--interactive` that defaults to false has already support for `--no-interactive` setting `interactive` arg to false.

zamotany:
@thymikee yeah but it doesn't work the same way, I've manually checked and unless I provide `=` it doesn't  detect correct value:
- `--no-interactive false` results in `undefiend`
- `--no-interactive true` results in `undefined`
Only `--no-interactive=true` sets it to `true`

thymikee:
The idea is to just run `--no-interactive` or `--interactive`, doesn't that work?

zamotany:
@thymikee Nope, was saying it's `undefined` when only `--no-interactive` was passed

thymikee:
Weird.

zamotany:
Screenshot 2019-07-23 at 12 53 01
@thymikee

packages/haul-core/src/server/NonInteractiveUI.ts
```diff
@@ -0,0 +1,33 @@
+import Runtime from '../runtime/Runtime';
+
+export default class NonInteractiveUserInterface {
```
thymikee: Maybe sharing a common interface between `InteractiveUI` and `NonInteractiveUI` makes sense? Unless TS infers it somehow already, then it's not needed

zamotany: I forgot to add `implemented UserInterface` as in `InteractiveUI`

thymikee:
Ah, it's probably because you name the flag as `no-interactive`. It should be called `interactive` for yargs, and it will figure out that when you prepend `no-`, it means false

zamotany:
Well then, in the end I still have to change that option to `interactive` and just mention `--no-interactive` in the docs.

thymikee:
Yep. And also you can mention somewhere at the top that:

> Note: for convenience, all boolean flags can be prepended with `no-` for `false` value, e.g. `--no-interactive` will set `interactive` to `false`.

PR #613, opened 2019-07-24
feat: replace @babel/register and manually transpile config modules on-demand

zamotany:


packages/haul-cli/src/commands/start.ts
```diff
@@ -87,19 +87,19 @@ export default function startCommand(runtime: Runtime) {
...
-          dev: argv.dev,
-          port: argv.port,
-          bundleMode: 'multi-bundle',
-          bundleTarget: 'server',
-          assetsDest: tempDir,
-          minify: argv.minify === undefined ? !argv.dev : argv.minify,
-        }
-      );
+        configPath
+      )(runtime, {
```
thymikee: why passing runtime again? it could be injected from the closure

zamotany: I could but it would be more problematic to type this, since I'll need to add another type: not only I would need `NormalizedProjectConfigBuilder` but also `NormalizedProjectConfigBuilderPartialApplied` os something which would not require `runtime` parameter. IMO it's unnecessary.

packages/haul-core/src/utils/importModule.ts
```diff
@@ -0,0 +1,194 @@
...
+    return nativeRequire(moduleFilename);
+  }
+
+  // Use exports from cache is available.
+  if (provided.cache[moduleFilename]) {
+    return provided.cache[moduleFilename].exports;
+  }
+
+  const moduleBody = fs.readFileSync(moduleFilename, 'utf8');
+  // Instantiating a new Module will setup some some properties, but won't
```
thymikee:
```suggestion
  // Instantiating a new Module will setup some properties, but won't
```

packages/haul-core/src/utils/importModule.ts
```diff
@@ -0,0 +1,194 @@
...
+  if (builtinModules.includes(moduleFilename)) {
+    return nativeRequire(moduleFilename);
+  }
+
+  // Use exports from cache is available.
+  if (provided.cache[moduleFilename]) {
+    return provided.cache[moduleFilename].exports;
+  }
+
+  const moduleBody = fs.readFileSync(moduleFilename, 'utf8');
```
thymikee: use try/catch for cleaner error?

PR #617, opened 2019-08-01
fix: add babel plugins to support Hermes

tido64:
Hermes doesn't understand ES6 (yet?) so we need the following babel plugins to support it.

### Summary

These plugins, and their options, were lifted from [`metro-react-native-babel-preset`](https://github.com/facebook/metro/blob/v0.55.0/packages/metro-react-native-babel-preset/src/configs/main.js). I'm not 100% whether this is exhaustive (as metro includes a lot more) but it works for us in a fairly large code base.

### Test plan

All tests should still pass.


tido64:
cc @acoates-ms, @thymikee, @zamotany

zamotany:
@tido64 How about we add those transforms or modify options only when `hermes: true` flag is passed to Babel preset:
```js
module.exports = {
  presets: ['module:@haul-bundler/babel-preset-react-native', { hermes: true }],
};
```
making if effectively opt-in?

Enabling those transforms will increate the bundle size by a significant margin, which we wanted to avoid with `@haul-bundler/babel-preset-react-native`. If we enable them by default, out babel preset will effectively be a clone of `metro-react-native-babel-preset`, in which case folks can just use Metro's preset.

zamotany:
Thanks @tido64 for this contribution. I'll do a release today EOD.

tido64:
@zamotany Thanks for the quick response! <3

PR #619, opened 2019-08-02
fix: hermes option read from wrong param

tido64:
### Summary

`hermes` option was read from the wrong parameter. This is now fixed.

### Test plan

All existing tests should still pass.


tido64:
@zamotany Sorry 😞

zamotany:
New version `@haul-bundler/babel-preset-react-native@0.10.1` is live on NPM.

PR #622, opened 2019-08-06
docs: add note about localhost/0.0.0.0 issue

michael-stern:
This commit updates the Configuration doc to include a note about
configuring the server to use host '0.0.0.0' if there is an issue
connecting to the dev server on localhost.

I experienced an issue where my react native app running on a ios device
could not connect to the development server when using the default
(localhost) server configuration. Changing the host to 0.0.0.0 fixed the
issue for me so others might find that advice useful as well.

### Summary

This pull request just adds a note to the docs that might help people solve issues connecting to the dev server from a device. 

### Test plan

n/a: docs only change


michael-stern:
@zamotany Thanks a ton for the help in Discord, I really appreciate it!

Let me know if you think this should go in a different section of the docs.

zamotany:
@michael-stern Thx for the PR 👍

Issue #621, opened 2019-08-06
makeConfig Does not use the port supplied in server


michael-stern:
## Environment

"react-native": "0.60.4",
"@haul-bundler/babel-preset-react-native": "0.10.1",
"@haul-bundler/cli": "0.10.0",
"@haul-bundler/preset-0.60": "0.10.0",

## Description

In my `haul.config.js` file I am calling `makeConfig` with a server object:
```
    server: {
        host: 'localhost',
        port: 8088,
    },
```
But when the haul server starts running it says `Packager server running on http://localhost:8081`
Changing the `host` in the `server` object works but it still ignores the `port`.

If I supply the port with `--port 8088` then it works as expected.

## Reproducible Demo

example haul config:
```
/* Copyright 2019 PolySign Inc.
import { withPolyfills, makeConfig } from '@haul-bundler/preset-0.60'

export default makeConfig({
    server: {
        host: '0.0.0.0',
        port: 8088,
    },
    bundles: {
        index: {
            entry: withPolyfills('./index.js'),
        },
    },
})
```

Then run haul with either `npm run start` (if you have a run script for that) or `npx haul start` and see that the port is still 8081.

PR #623, opened 2019-08-07
fix: port setting in haul.config.js not used

zamotany:
#621

### DIFF
--- a/packages/haul-cli/src/commands/start.ts
+++ b/packages/haul-cli/src/commands/start.ts
@@ -1,6 +1,5 @@
 import yargs from 'yargs';
 import {
-  DEFAULT_PORT,
   DEFAULT_CONFIG_FILENAME,
   INTERACTIVE_MODE_DEFAULT,
   getProjectConfigPath,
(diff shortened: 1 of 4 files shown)


packages/haul-cli/src/commands/start.ts
```diff
@@ -22,7 +21,6 @@ export default function startCommand(runtime: Runtime) {
     builder: {
       port: {
         description: 'Port to run your webpack server',
-        default: DEFAULT_PORT,
```
thymikee: why removing from here? I guess it's still normalized later, but this feels like a proper place for a default port

zamotany: The thing is if I add it here the `port` from `haul.config.js` won't be picked up. Also we use `port` from the normalised config and there's a fallback to `DEFAULT_PORT` there as well:
```
port: env.port || get(projectConfig, 'server.port', DEFAULT_PORT),
```

thymikee: Ah yea that's a fair call

PR #624, opened 2019-08-08
fix: Speed up source map generation by only stringiying at the end

FuegoFro:
Currently when generating source maps, we stringiyf the in-progress
indexMap once for every. single. file. even though we only care about
the final result (and indeed overwrite our previous stringification each
file). This was likely a typo and intended to be outside the forEach, so
this change moves the stringification to after we're fully done building
up the indexMap. This dramatically speeds up source map generation since
we're doing significantly less work with fewer allocations and GCs.


zamotany:
@FuegoFro Thank you this PR 👍

FuegoFro:
Thank you for the review+merge! 😀

PR #625, opened 2019-08-09
chore: update dependencies

zamotany:
- most of the dependencies was updated to newest version except for babel
- babel dependencies were not updates, because updating them to latest version (minor bump) is causing crashes in `importModule` - this need more research etc
- manual tests after dependency were made to ensure nothing got broken


package.json
```diff
@@ -31,18 +31,18 @@
...
     "@babel/preset-env": "^7.4.3",
     "@babel/preset-typescript": "^7.3.3",
-    "@callstack/eslint-config": "^4.2.0",
-    "@types/jest": "^24.0.11",
-    "@types/webpack": "^4.4.29",
+    "@callstack/eslint-config": "^7.0.0",
+    "@types/jest": "^24.0.17",
     "@types/mkdirp": "^0.5.2",
-    "@types/node": "^11.13.4",
+    "@types/node": "^12.7.0",
```
thymikee: FYI according to @Esemesek if we target Node 10, we should be using `@types/node@^10.0.0` to make sure we use compatible APIs

packages/haul-core-legacy/package.json
```diff
@@ -26,11 +26,10 @@
   "dependencies": {
     "mime-types": "^2.1.24",
     "mkdirp": "^0.5.1",
-    "ws": "^2.2.2",
+    "ws": "^6.2.1",
```
thymikee: good to know this is not breaking anything

packages/haul-core/src/server/setupCompilerRoutes.ts
```diff
@@ -78,8 +76,8 @@ export default function setupCompilerRoutes(
         );
 
         if (!hasRunAdbReverse && platform === 'android') {
-          await runAdbReverse(runtime, port);
           hasRunAdbReverse = true;
```
thymikee: it hasn't yet :D maybe worth changing the variable name if this is intended?

packages/haul-ram-bundle-webpack-plugin/package.json
```diff
@@ -12,11 +12,11 @@
   "license": "MIT",
   "bugs": "https://github.com/callstack/haul/issues",
   "peerDependencies": {
-    "@haul-bundler/core": "^0.10.0",
-    "metro": "^0.53.1",
-    "terser": "^3.17.0",
-    "webpack": "^4.30.0",
-    "webpack-sources": "^1.3.0"
+    "@haul-bundler/core": "0.10.0",
+    "metro": ">= 0.53.1",
```
thymikee: quite permissive, maybe worth to set a top limit to 0.55?

packages/haul-core-legacy/package.json
zamotany: I test it manually and it's fine.

packages/haul-core/src/server/setupCompilerRoutes.ts
zamotany: ESLint: possible race condition if `hasRunAdbReverse` is changed after async operation.

packages/haul-ram-bundle-webpack-plugin/package.json
zamotany: Technically we should make it `0.51.1` and above since this one ships with RN 0.59

packages/haul-core/src/server/setupCompilerRoutes.ts
thymikee: I'm fine as long as it makes sense in this scope (I didn't dive into how `hasRunAdbReverse` is used), just wanted to point it out

packages/haul-ram-bundle-webpack-plugin/package.json
zamotany: @thymikee WDYT about `>=0.51.1 < 1`?

thymikee: depends on the APIs Haul use and how stable are they, but if they sit there for a while, I'm good

packages/haul-core/src/server/setupCompilerRoutes.ts
zamotany: `hasRunAdbReverse` is only to make sure `runAdbReverse` is run only once. If it fails we print error and let the user figure out what to do. We definitely don't want to run `runAdbReverse` multiple times. 

If `hasRunAdbReverse` is set after the `runAdbReverse` there is a chance that it will be called 2 times or more, since the next request will have a old `hasRunAdbReverse` value - not updated yet.

packages/haul-ram-bundle-webpack-plugin/package.json
zamotany: Well we've been using `RamBundleParser` from `0.51.1` and newer without any issues.

PR #633, opened 2019-08-27
feat: external bundles

zamotany:
Adds a way to use pre-built (DLL) bundle when using `multi-bundle` command.

### TODO

- [x] add support for `bundlePath` and `manifestPath` to `multi-bundle` command
- [x] add support for `bundlePath` and `manifestPath` to `packager server
- [x] serve source maps for pre-built bundle from packager server
- [x] add `copyBundle` option to automatically copy files to target directory
- [x] add integration test
- [x] monorepo fixture
- [x] monorepo integration test
- [x] add documentation
- [x] properly handle and resolve assets from pre-built bundle


docs/Configuration.md
```diff
@@ -376,3 +381,83 @@ class RootComponent extends React.Component {
 
 AppRegistry.registerComponent('MyApp', () => RootComponent);
 ```
+
+### External (pre-built) bundles
+
+Check out [monorepo fixture](../fixtures/monorepo) for a complete an example.
+
+To use external bundles, first built the bundles that will be considered pre-built. You only need a single bundle config:
```
thymikee:
```suggestion
To use external bundles, first build the bundles that will be considered pre-built. You only need a single bundle config:
```

docs/Configuration.md
```diff
@@ -376,3 +381,83 @@ class RootComponent extends React.Component {
...
+
+export default makeConfig({
+  bundles: {
+    index: {
+      entry: withPolyfills('./src/host.js'),
+      dependsOn: ['base_dll'],
+    },
+    // Let's assume base-dll bundle is a Node module.
+    base_dll: ({ platform, bundleTarget, dev }) => {
+      // For a pre-built bundle, both development and production bundle have to be preset.
```
thymikee:
```suggestion
      // For a pre-built bundle, both development and production bundle have to be present.
```

fixtures/monorepo/host/BundleRegistry.js
```diff
@@ -0,0 +1,72 @@
+import EventEmitter from 'react-native/Libraries/vendor/emitter/EventEmitter';
+import BatchedBridge from 'react-native/Libraries/BatchedBridge/BatchedBridge';
+
+class BundleRegistry extends EventEmitter {
```
thymikee: Is it possible to offload this logic to Haul?

docs/Configuration.md
```diff
@@ -376,3 +381,83 @@ class RootComponent extends React.Component {
...
+      entry,
+      dll: true,
+      // Packager server will use development bundle, so it should be built as a plain JS bundle.
+      type: dev ? 'basic-bundle' : 'indexed-ram-bundle',
+    }),
+  },
+});
+```
+
+Use `--skip-host-check` option when building pre-built bundles:
```
thymikee: Is it possible to automatically skip host check for pre-built bundles? This flag seems quite unintuitive

packages/haul-core-legacy/src/compiler/worker/runWebpackCompiler.js
```diff
@@ -50,7 +53,101 @@ module.exports = async function runWebpackCompiler({
...
+      const bundleFilename = getBundleFilename(
+        {
+          ...configOptions,
+          platform,
+        },
+        projectConfig.templates,
+        bundleConfig
+      );
+
+      // TODO: assets
```
thymikee: todo after the PR?

packages/haul-core-legacy/src/compiler/worker/runWebpackCompiler.js
```diff
@@ -89,10 +186,6 @@ module.exports = async function runWebpackCompiler({
 
   const compiler = webpack(apps);
 
-  // As of Webpack 4.12.0, |outputFileSystem| must be set since there is no
-  // fallback despite the documentation stating otherwise.
-  compiler.outputFileSystem = fs;
```
thymikee: was this fixed in some recent version? if so, we need to bump peer dep

packages/haul-core/src/config/types.ts
```diff
@@ -47,6 +47,11 @@ export type BundleConfig = {
   >;
   hasteOptions?: any;
   transform?: WebpackConfigTransform;
+  /* external bundles */
```
thymikee: instead of a comment, how about creating a new type name `ExternalBundleConfig` and spread/join it here

packages/haul-core/src/config/types.ts
```diff
@@ -55,9 +60,33 @@ export type TemplatesConfig = {
...
+  Overwrite<
+    Pick<
+      DeepNonNullable<BundleConfig>,
+      Exclude<
+        keyof BundleConfig,
+        | 'transform'
+        | 'copyBundle'
+        | 'bundlePath'
+        | 'manifestPath'
+        | 'assetsPath'
```
thymikee: thee could be taken from `ExternalBundleConfig` type if created

fixtures/monorepo/host/BundleRegistry.js
zamotany: It is, but I don't want to do it right now, but later, once we have everything in RN that we need to creating native module for multi-bundle support

docs/Configuration.md
zamotany: Not sure how to reliably detect when to run this check. IMO we either leave this flag or remove the check.

packages/haul-core-legacy/src/compiler/worker/runWebpackCompiler.js
zamotany: Leftover

zamotany:
```suggestion
```

packages/haul-core-legacy/src/compiler/worker/runWebpackCompiler.js
zamotany: we don't use custom filesystem, the fs here was actually the leftover from where we were using `memory-fs`, but now it was pointing to `fs` module from Node.

docs/Configuration.md
thymikee: If the check may cover some of the common fail points, let's leave it as is

PR #640, opened 2019-09-13
fix: only add plugin-transform-react-jsx-source in dev

tido64:
### Summary

Conditionally add `@babel/plugin-transform-react-jsx-source` so we don't get `__source` injected into production bundles, and reduce the bundle size.

### Test plan

All current tests should pass.


tido64:
cc @zamotany @thymikee

PR #642, opened 2019-09-17
chore: add e2e bundling tests

zamotany:


zamotany:
@thymikee Some context for you:
- I plan to move some tests from `integration_tests` to `e2e`, where some commands will be spawned in subprocess for different fixtures just to make sure they work
- In `integration_tests` there will be tests for a single fixture project and won't be spawned in a subprocess but called directly in the same context as the test (+ some mocking), so that we get proper coverage

e2e/react_native_windows_vnext/__tests__/bundling.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  validateHostBundle,
+  validateAppBundle,
+} from '../../utils/bundles';
+
+const PROJECT_FIXTURE = path.join(
+  __dirname,
+  '../../../fixtures/react_native_windows_vnext'
+);
+
+describe('react_native_windows_vnext', () => {
```
thymikee: these tests for windows current and vnext are quite repetitive, how about extract helpers and pass paths etc accordingly?

PR #645, opened 2019-09-20
chore: improve e2e tests

zamotany:


e2e/react_native_0_60x/__tests__/start.test.ts
```diff
@@ -0,0 +1,3 @@
+import createStartTestSuite from '../../utils/createStartTestSuite';
+
+createStartTestSuite('react_native_with_haul_0_60x', 8082);
```
thymikee: I wonder, maybe instead of passing different ports here and there, you could use `process.env.JEST_WORKER_ID`: https://github.com/facebook/jest/tree/master/packages/jest-worker#end. It's bound to the worker, so there if there's no multiple server running during a single test file concurrently (likely not) then you should be fine to set the port to `8080 + process.env.JEST_WORKER_ID` everywhere

PR #649, opened 2019-10-03
chore: e2e windows tests

zamotany:
TODO:
- [x] switch to local Haul instead of packages from NPM


babel.config.js
```diff
@@ -15,4 +15,9 @@ module.exports = {
     '@babel/plugin-transform-flow-strip-types',
     '@babel/plugin-transform-react-jsx',
   ],
+  env: {
+    development: {
+      ignore: [/__tests__/],
```
thymikee: Do you import something from `__tests__`? Babel shouldn't run otherwise I guess

fixtures/react_native_windows_e2e_app/__tests__/appium.test.js
```diff
@@ -0,0 +1,63 @@
...
+  expect(await getByTestId('app0')).toBeDefined();
+  expect(await (await getByTestId('app0LoadTime')).text()).toMatch(/Load time: [1-9]+ ms/);
+
+  (await getByTestId('navigateToapp1')).click();
+  await driver.sleep(1000);
+
+  expect(await getByTestId('app1')).toBeDefined();
+  expect(await (await getByTestId('app1LoadTime')).text()).toMatch(/Load time: [1-9]+ ms/);
+  
+  await driver.sleep(20000);
```
thymikee: that's a long time, no way to detect changes?

babel.config.js
zamotany: I don't want to transpile files in `__tests__` when building the packages, but I do want files in `__tests__` to be transpiled when testing. I don't want to specify `--ignore` flag when running babel, due to issues on windows.

fixtures/react_native_windows_e2e_app/windows/RNWTestApp.sln
```diff
@@ -0,0 +1,140 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
thymikee: Do we need a solution file checked into source control? I don't see any absolute paths, so that may be fine for convenience

zamotany: Yes we do.

fixtures/react_native_windows_e2e_app/windows/RNWTestApp.sln
```diff
@@ -0,0 +1,140 @@
...
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29324.140
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "RNWTestApp", "RNWTestApp\RNWTestApp.csproj", "{72EE9031-61F5-4539-AEE8-F3AD644507B5}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Folly", "..\node_modules\react-native-windows\Folly\Folly.vcxproj", "{A990658C-CE31-4BCC-976F-0FC6B1AF693D}"
```
matthargett: I'd recommend changing these relative paths to $SolutionDir) prefixed paths for less headaches in the future.

PR #655, opened 2019-10-07
docs: update react-native-windows recipe

zamotany:
#650

### DIFF
--- a/docs/Recipes.md
+++ b/docs/Recipes.md
@@ -157,7 +157,7 @@ export default makeConfig({
   platforms: ['windows', 'ios', 'android'], // or ['windows'] if you only use windows platform
   bundles: {
     index: {
-      entry: entry: withPolyfills('./index.js', {
+      entry: withPolyfills('./index.js', {
         initializeCoreLocation: 'node_modules/react-native-windows/Libraries/Core/InitializeCore.js'
       }),
       providesModuleNodeModules: ['react-native', 'react-native-windows'],
--- a/docs/Recipes.md
+++ b/docs/Recipes.md
@@ -125,13 +125,16 @@ export default makeConfig({
 
 ## Use Haul with `react-native-windows`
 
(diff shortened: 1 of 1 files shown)


docs/Recipes.md
```diff
@@ -142,6 +145,35 @@ export default makeConfig({
 });
 ```
 
+### With React Native 0.60 and newer
+
+React Native 0.60 removed Haste imports, so now in order to use out-of-tree platforms, you need to modify `InitializeCore` location and alias `react-native` to your out-of-tree library.
```
thymikee: Technically they removed _some_ of the imports, all of them were removed in 0.61. Not sure if it makes sense to make it public :D

docs/Recipes.md
```diff
@@ -142,6 +145,35 @@ export default makeConfig({
...
+
+Update your `haul.config.js` like so:
+```js
+import { makeConfig, withPolyfills } from "@haul-bundler/preset-0.60";
+
+export default makeConfig({
+  platforms: ['windows', 'ios', 'android'], // or ['windows'] if you only use windows platform
+  bundles: {
+    index: {
+      entry: entry: withPolyfills('./index.js', {
```
thymikee:
```suggestion
      entry: withPolyfills('./index.js', {
```

docs/Recipes.md
```diff
@@ -142,6 +145,35 @@ export default makeConfig({
...
+        initializeCoreLocation: 'node_modules/react-native-windows/Libraries/Core/InitializeCore.js'
```
thymikee: wouldn't `require.resolve` be nicer?

zamotany: Well, it wouldn't work. `initializeCoreLocation` is joined with `root` option, which you can omit as in here and it will be set to `process.cwd()`. I'm not entirely sure why is that, but for sure doing `require.resolve` inside `withPolyfills` will not always work and we don't want to require users to specify `initializeCoreLocation` everytime with `require.resolve`.

thymikee: Get it. Paths are tricky :|

PR #656, opened 2019-10-10
chore: setup code coverage report & check from unit and e2e tests

zamotany:


babel.config.js
```diff
@@ -16,8 +16,14 @@ module.exports = {
     '@babel/plugin-transform-react-jsx',
   ],
   env: {
+    // Building for local development and E2E tests
     development: {
       ignore: [/__tests__/],
+      plugins: ['babel-plugin-istanbul'],
```
thymikee: how about putting only in "test" environment?

zamotany: Would work only for unit tests, for e2e we have to use built version of packages and they have to be built with `babel-plugin-istanbul`.

PR #664, opened 2019-10-14
feat: export functions for getting a list of plugins and utilities

zamotany:
#660

### DIFF
--- a/packages/haul-babel-preset-react-native/src/index.ts
+++ b/packages/haul-babel-preset-react-native/src/index.ts
@@ -1,47 +1,135 @@
 import path from 'path';
 
-const defaultPlugins = [
-  [require('@babel/plugin-proposal-class-properties'), { loose: true }],
-  [require('@babel/plugin-proposal-optional-catch-binding')],
-  [require('@babel/plugin-syntax-dynamic-import')],
-  [require('@babel/plugin-syntax-export-default-from')],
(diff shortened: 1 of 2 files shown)


packages/haul-babel-preset-react-native/src/index.ts
```diff
@@ -1,49 +1,137 @@
...
+
+export function getDefaultPrePlugins(): PluginSpec[] {
+  return [
+    {
+      // The flow strip types plugin must go BEFORE class properties!
+      name: '@babel/plugin-transform-flow-strip-types',
+    },
+    {
+      name: '@babel/plugin-proposal-class-properties',
+      options: { loose: true },
```
thymikee: Wouldn't it be easier to conform to Babel configuration scheme instead of creating new and mapping it back to Babel? So PluginSpec would be something like `string | [string, object]`.

zamotany: I think it's easier to use object or `[string, object]` tuple for consuming `@haul-bundler/babel-preset-react-native` as a base for custom preset.

thymikee: I'm fine as long as it's documented. Just want you to keep an eye on limiting the API surface, it's a pain to learn different config schemes for different tools and we should avoid it.

PR #665, opened 2019-10-14
feat: install previous version of preset if the target one is not available

zamotany:
#663


e2e/react_native_0_61x/__tests__/init.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  cleanup(path.join(TEST_PROJECT_DIR, '..'));
+  mkdirp.sync(path.join(TEST_PROJECT_DIR, '..'));
+});
+afterEach(() => cleanup(path.join(TEST_PROJECT_DIR, '..')));
+
+test('init command on a fresh react-native 0.61 project', done => {
+  run(`${yarnCommand} init -y`, path.join(TEST_PROJECT_DIR, '..'));
+  run(`${yarnCommand} add react-native-cli`, path.join(TEST_PROJECT_DIR, '..'));
+  run(
+    `${yarnCommand} react-native init testapp --version 0.61.0`,
```
thymikee: We advise using `npx` now, which uses new `init`. So essentially a single command, without yarn init, would suffice:
```
npx react-native init testapp --version 0.61.0
```

packages/haul-cli/src/commands/init.ts
```diff
@@ -307,11 +309,36 @@ async function addHaulScript(progress: Ora, cwd: string): Promise<string> {
...
+  // Stop searching on 0.59 - there's no preset below 0.59.
+  if (targetHaulPreset.includes('0.59')) {
+    return targetHaulPreset;
+  }
+
+  try {
+    await npmFetch(targetHaulPreset);
+    return targetHaulPreset;
+  } catch (error) {
+    if (error.statusCode === 404) {
```
thymikee: I think it would be nice to log a message about falling back to previous preset, wdyt?

PR #668, opened 2019-10-28
chore: use single tsc build and move types to monorepo root

zamotany:


package.json
```diff
@@ -5,19 +5,22 @@
...
-    "watch": "lerna exec --parallel \"babel src --out-dir build --extensions .js,.ts,.tsx --source-maps --delete-dir-on-start --config-file=../../babel.config.js --watch\"",
-    "build:def": "lerna exec --ignore @haul-bundler/core-legacy \"tsc --emitDeclarationOnly -p ./tsconfig.json --outDir ./build\"",
-    "prepare": "yarn build:source && yarn build:def",
-    "typecheck": "tsc --noEmit",
     "test": "jest",
+    "clean": "rm -rf ./packages/*/build ./packages/*/tsconfig.tsbuildinfo",
+    "build:source": "lerna exec --parallel \"babel src --out-dir build --extensions .js,.ts,.tsx --source-maps --config-file=../../babel.config.js\"",
+    "watch:source": "lerna exec --parallel \"babel src --out-dir build --extensions .js,.ts,.tsx --source-maps --config-file=../../babel.config.js --watch\"",
+    "build:ts": "node ./scripts/build_ts.js",
+    "watch:ts": "node ./scripts/build_ts.js --watch",
```
thymikee: In my experience watching TS and Babel at the same time has suboptimal DX, since you need to run both in different terminale panes/tabs. Ideas around that? I guess we need both TS and Babel for now, because not all packages are converted. Would be cool to also mention the workflow in CONTRIBUTING.md file.

PR #670, opened 2019-10-30
fix: transform spread operator for Hermes

tido64:
### Summary

Hermes doesn't support spread operator. Added the necessary plugins to transform them to ES5 syntax.

### Test plan

All tests should pass.


tido64:
cc @zamotany, @thymikee

PR #674, opened 2019-11-05
fix: additional plugins for Hermes

tido64:
### Summary

Depending on your setup and usage (e.g. macOS vs Windows, bundle vs. dev server), Hermes won't be able to execute your bundle without these additional plugins.

### Test plan

All current tests should pass.


tido64:
cc @zamotany, @thymikee

Issue #678, opened 2019-11-15
App fails to start when used with vscode react native debugger: "NodeError: Invalid URL"


hedgepigdaniel:
## Environment

Haul: 0.14.2
vscode-react-native-debugger: 0.13.1

## Description

`debuggerWorker.js` crashes if the bundle URL is a local filepath instead of an HTTP URL (in the [VSCode react native debugger](https://github.com/microsoft/vscode-react-native), the URL is a local file path).

See https://github.com/microsoft/vscode-react-native/issues/1164 for more detailed description

I think the problem was introduced by https://github.com/callstack/haul/pull/641

zamotany:
The path from `vscode-react-native` is missing a protocol `file://` and thus it's invalid. I'd suggest to fix that in upsteam (`vscode-react-native`), instead of hacking it in `debuggerWorker.js`

zamotany:
@hedgepigdaniel This should be fixed now.

PR #680, opened 2019-11-20
fix: assign location.host only if self is available in debuggerWorker.js

zamotany:
#678


zamotany:
Unblocks https://github.com/microsoft/vscode-react-native/pull/1166

PR #681, opened 2019-11-21
Yarn workspace test fixture

matthargett:
### Summary

This is a test fixture for 0.59 in a yarn workspace with proper hoisting. It reproduces an issue we have in jest, but I'd also like to see this real-world scenario covered in CI.

### Test plan

Ran `yarn test` from the root. I get some failures that appear to be due to my Windows environment and are unrelated to the changes. I'd love help figuring that out, as the tests should be passing in


thymikee:
If Jest is used in a monorepo and deps are hoisted, you need to pass custom `rootDir` pointing to the directory where it's hoisted. I know it's bad and we'd like to fix it as some point, but for now it should unblock you. Here's the diff (I had to dedupe RN as well):

```diff
diff --git a/fixtures/monorepo-yarn-workspace/packages/app0/__tests__/App.test.js b/fixtures/monorepo-yarn-workspace/packages/app0/__tests__/App.test.js
index 520b83c..5922637 100644
--- a/fixtures/monorepo-yarn-workspace/packages/app0/__tests__/App.test.js
+++ b/fixtures/monorepo-yarn-workspace/packages/app0/__tests__/App.test.js
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { mount } from 'enzyme';
-import App from '..';
+import App from '../App';

 let subject;
 afterEach(() => {
diff --git a/fixtures/monorepo-yarn-workspace/packages/app0/jest.config.js b/fixtures/monorepo-yarn-workspace/packages/app0/jest.config.js
index 5591324..2a0fa0a 100644
--- a/fixtures/monorepo-yarn-workspace/packages/app0/jest.config.js
+++ b/fixtures/monorepo-yarn-workspace/packages/app0/jest.config.js
@@ -11,4 +11,5 @@ module.exports = {
     '.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga|dds|ogg|json.resource)$':
       require.resolve('react-native/jest/assetFileTransformer.js'),
   },
+  rootDir: "../../"
 };
diff --git a/fixtures/monorepo-yarn-workspace/packages/app0/package.json b/fixtures/monorepo-yarn-workspace/packages/app0/package.json
index 77c5c8f..62da0bf 100644
--- a/fixtures/monorepo-yarn-workspace/packages/app0/package.json
+++ b/fixtures/monorepo-yarn-workspace/packages/app0/package.json
@@ -16,7 +16,7 @@
   },
   "dependencies": {
     "react": "16.8.3",
-    "react-native": "0.59.10",
+    "react-native": "callstack/react-native#feat/apennine-0.59",
     "react-native-gesture-handler": "^1.3.0",
     "react-navigation": "^3.11.0",
     "base-dll": "*"
diff --git a/fixtures/monorepo-yarn-workspace/packages/base-dll/package.json b/fixtures/monorepo-yarn-workspace/packages/base-dll/package.json
index 989d195..ae07b94 100644
--- a/fixtures/monorepo-yarn-workspace/packages/base-dll/package.json
+++ b/fixtures/monorepo-yarn-workspace/packages/base-dll/package.json
@@ -16,7 +16,7 @@
   },
   "dependencies": {
     "react": "16.8.3",
-    "react-native": "0.59.10",
+    "react-native": "callstack/react-native#feat/apennine-0.59",
     "react-native-gesture-handler": "^1.3.0",
     "react-navigation": "^3.11.0"
   }
```

After that I run into this:

image

But IIRC this is expected, since we're running in a `node` environment, and Enzyme expects `jsdom` when using `mount` (with `react-dom` as well I guess).

matthargett:
Talking with the affected  team, they tried rootDir and ran into another problem when there is a platform override of a core component in the haste map, and that out-of-tree platform package is hoisted. I'm going to add another fixture to this PR based on the existing react-native-windows fixture to illustrate this part of the problem.

I've added your diffs above in the meantime -- thanks!

joeblynch:
@thymikee Thanks for looking into this! Setting `rootDir` resolved that issue, however we ran into another one afterwards, where we hit this error:
```
Test suite failed to run

    Cannot find module '../Libraries/Image/Image' from 'setup.js'
      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:259:17)
      at Object.<anonymous> (node_modules/react-native/jest/setup.js:57:5)
```

Have you seen this before? At first we thought it was still related to using a monorepo, but we were able to reproduce the same issue without a monorepo being part of the equation. Best I can tell, it happens when adding a component from `react-native-windows` to the test, due to including an out of tree platform. I have the test that replicates the issue in @matthargett's fork [here](https://github.com/matthargett/haul/tree/jest-haste-config-fixture/fixtures/react_native_windows_current_jest_haste).

thymikee:
Looking at the config: https://github.com/matthargett/haul/blob/197449d0a776bd26a64c3478feefe99454d42881/fixtures/react_native_windows_current_jest_haste/jest.config.js#L14-L19 seems like you tell Jest to only look for a single platform but in both `react-native` and `react-native-windows`. It tries to resolve `Libraries/Image/Image.windows.js` from `react-native` which is not there. Have you tried swapping the order of `providesModuleNodeModules` to be `['react-native-windows', 'react-native']`? I'm not that familiar with Haste options unfortunately, but that would be my guess (haven't run the code yet, hopefully will find the time tomorrow or so)

matthargett:
@thymikee We tried swapping the entries, it had no effect on the output. If you need more to estimate the work to fix the jest problem, let us know.  thanks for all your suggestions so far!

PR #697, opened 2020-02-05
fix: use `bundleId` instead of `bundleName` for multibundle

simka:

### DIFF
--- a/packages/haul-core/src/config/types.ts
+++ b/packages/haul-core/src/config/types.ts
@@ -101,10 +101,15 @@ export type BundleConfigBuilder = (
   runtime: Runtime
 ) => BundleConfig;
 
+export type FeaturesConfig = {
+  multibundle?: 1 | 2;
+};
+
 export type ProjectConfig = {
   server?: ServerConfig;
   platforms?: string[];
   templates?: TemplatesConfig;
+  features?: FeaturesConfig;
   bundles: { [bundleName: string]: BundleConfigBuilder | BundleConfig };
 };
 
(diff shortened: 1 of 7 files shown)


zamotany:
Looks good, but we need to make it backwards-compatible. Let's add:
```
// haul.config.js
export default makeConfig({
  features: {
    multiBundle: 2,
  },
  bundles: {
    index: { /* ... */ }
  }
});
```
The default for `features.multiBundle` should be 1 (if not specified).
Then here:
https://github.com/callstack/haul/blob/71ca671237fad3fb552f187d4de955e4464481e8/packages/haul-core/src/preset/makeConfigFactory.ts#L252-L256
depending on `features.multiBundle` we would pass either `bundleName` or `bundleIdsMap[bundleName]`

simka:
> Then here:
> https://github.com/callstack/haul/blob/71ca671237fad3fb552f187d4de955e4464481e8/packages/haul-core/src/preset/makeConfigFactory.ts#L252-L256
> 
> 
> depending on `features.multiBundle` we would pass either `bundleName` or `bundleIdsMap[bundleName]`

before this PR there was no third argument, so shouldn't it be either `undefined` or `bundleIdsMap[bundleName]` and then in `haul-ram-bundle-webpack-plugin/runtime/bootstrap.js` we could use `bundleId` if available or `bundleName` if not?

zamotany:
> before this PR there was no third argument, so shouldn't it be either undefined or bundleIdsMap[bundleName] and then in haul-ram-bundle-webpack-plugin/runtime/bootstrap.js we could use bundleId if available or bundleName if not?

You're right there wasn't 3rd argument and the `bundleName` passed to `runtime/bootstrap.js` was taken from 
https://github.com/callstack/haul/blob/3ae2a9465fe8519a237b795bcfab4d19e3627bc5/packages/haul-ram-bundle-webpack-plugin/src/WebpackRamBundlePlugin.ts#L98-L103

I suggest to remove this logic above and accept `bundleId: number | string` in the `WebpackRamBundlePlugin` constructor and figure out what value to pass in `makeConfigFactory.ts`

PR #716, opened 2020-04-03
fix: live reload by resolving watchers only if they weren't notified

zamotany:


packages/haul-core/src/server/setupLiveReload.ts
```diff
@@ -35,11 +36,13 @@ export default function setupLiveReload(
...
          */
-        const watcher = request.raw;
+        const watcher = Object.assign(request.raw, { notified: false });
         watchers.push(watcher);
         watcher.req.on('close', () => {
           watchers.splice(watchers.indexOf(watcher), 1);
-          resolve('OK');
+          if (!watcher.notified) {
+            resolve('OK');
+          }
```
tido64, line 45: I think this would also work:
```js
        const index = watchers.indexOf(watcher);
        if (index >= 0) {
          watchers.splice(index, 1);
          resolve('OK');
        }
```

statm:
Any update on this? Looks like build failed because of an extra `--force` flag at https://github.com/callstack/haul/blob/master/azure-pipelines.yml#L39

packages/haul-core/src/server/setupLiveReload.ts
sorrly: I am interested in this project!,
When is this bug fixed on main branch??

zamotany: It will be once I merge the PR and make a release.

zamotany:
@statm 
The E2E test on Azure Pipelines has been failing for quite some time. The `--force` flag is not the real issue, but the React Native Windows not compiling due to missing component. I've little experience in working with Windows and limited bandwidth to fix the issue.

zamotany:
Published under:
 - @haul-bundler/cli@0.18.1
 - @haul-bundler/core@0.18.1

PR #717, opened 2020-04-10
fix: add `--skip-host-check` flag to `start` command

statm:
### Summary

For `haul bundle` command, there's a [restriction](https://github.com/callstack/haul/blob/22efe80fc7e0d4f50c046d5dd767a15c32428cf1/packages/haul-cli/src/commands/shared/prepareWebpackConfig.ts#L53) that webpack config must contain `index` or `main` entry. This restriction can be lifted for `haul multi-bundle` via [`--skip-host-check`](https://github.com/callstack/haul/blob/master/packages/haul-cli/src/commands/multiBundle.ts#L137) flag, but cannot be bypassed when running `haul start`.

This PR ports `--skip-host-check` flag to `haul start` so multi-bundle settings without `index` nor `main` entry can be served.

### DIFF
--- a/packages/haul-cli/src/commands/start.ts
+++ b/packages/haul-cli/src/commands/start.ts
@@ -53,6 +53,11 @@ export default function startCommand(runtime: Runtime) {
         default: 'false',
         type: 'string',
       },
+      'skip-host-check': {
+        description: 'Skips check for "index" or "host" bundle in Haul config',
+        default: 'false',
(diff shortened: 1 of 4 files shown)


statm:
cc @zamotany

zamotany:
Released to NPM, you can update `@haul-bunlder/cli` to `0.19.0`.

PR #721, opened 2020-04-20
chore: remove haul inspector

NoemiRozpara:
### Summary

Remove haul-inspector and haul-inspector-events without breaking e.g. node inspector

### Test plan

- [x] Run server with node inspector
Succesfully ran `yarn haul bundle --platform ios --dev false --node-inspector=wait`


thymikee:
Awesome! Could you please add a rationale behind that? The PR is removing 2 packages, so we owe our users an explanation :)

zamotany:
The reasoning behind removing haul inspector is that, it doesn't give much benefit over doing standard Node debugging and/or outputting logs to a file. In near future we plan to improve logging, including more debug logs, which will be more verbose, so that the users can run haul and investigate those verbose logs in a file.

PR #729, opened 2020-04-27
Feat/support bundle options

TMaszko:


e2e/react_native_0_60x_ts/__tests__/bundleOptions.android.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+async function startServerAsync(): Promise<Instance> {
+  return new Promise((resolve, reject) => {
+    const done = () => {
+      resolve(instance);
+    };
+    done.fail = reject;
+    const instance = startServer(PORT, TEST_PROJECT_DIR, undefined, done);
+  });
+}
```
zamotany, line 85: You could move those functions to some util file, since they are duplicated. My suggestion is to move `fetchBundle` to `e2e/utils/bundle.ts` and `startServerAsync` to `e2e/utils/server.ts`

e2e/utils/validators.ts
```diff
@@ -5,6 +5,7 @@ import fetch from 'node-fetch';
 type ValidationOptions = {
   isIndexedRAMBundle?: boolean;
   platform?: string;
+  log?: boolean;
```
zamotany, line 8: Is this option used anywhere?

packages/haul-core/src/server/Server.ts
```diff
@@ -202,6 +202,10 @@ export default class Server {
       port,
       bundleNames: this.options.bundleNames,
       platforms: this.options.platforms,
+      cliBundleOptions: {
+        minify: this.options.minify,
+        dev: this.options.dev,
+      },
```
zamotany, line 208: Those should not be set-able from the CLI.

e2e/utils/validators.ts
TMaszko: No, I forgot about it when I was removing whole log option implementation

packages/haul-core/src/server/Server.ts
zamotany: Nvm, I thought `start` command doesn't have `--dev/--minify`

e2e/react_native_0_60x_ts/__tests__/bundleOptions.android.test.ts
TMaszko: Thanks good idea !

PR #731, opened 2020-04-29
Added warn msg to terminal output when user changed bundle options

TMaszko:


packages/haul-core/src/server/setupCompilerRoutes.ts
```diff
@@ -110,9 +110,10 @@ export default function setupCompilerRoutes(
         );
 
         if (isUserChangedOptions) {
-          return h
-            .response('To see the changes you need to restart the haul server')
-            .code(501);
+          const warnMsg =
+            'To see the changes you need to restart the haul server';
```
zamotany, line 114:
```suggestion
            'Changing query params after the bundle has been created is not supported. To see the changes you need to restart the Haul server.';
```

PR #689, opened 2020-01-16
feat: move tasks to workers

NoemiRozpara:
### Summary

RAM bundle build time boost using `jest-workers`. Number of workers can be con figured using `--maxWorkers` option, e.g.
```
yarn haul ram-bundle --platform ios --dev false  --max-workers 7
```

### Test plan

Proper test was added to the fixtures 

### To do 
- [x] Think about limiting workers number
- [x] Move setting inferred `maxWorkers` value to `makeConfigFactory`
- [x] Add `--max-workers` documentation
- [x] Add documentation about `--progress` performance implications
- [x] Cleanup `babelWorkerLoader`
- [x] Switch to ESM in `babelWorkerLoader/*`
- [x] Remove commented out code from `preset-*/src/defaultConfig.ts`
- [x] Revert changes to `IndexRamBundle` and FileRamBundle` source map's line generation

PR #735, opened 2020-05-06
fix: Fix incorrect sourcemap overwrite during transform

statm:
### Summary

In our project we use a different loader setting for TypeScript files from what's recommended in the [Recipes](https://github.com/callstack/haul/blob/master/docs/Recipes.md#with-ts-compiler-tsc).
TS files are loaded by `ts-loader` **then** a `babelWorkerLoader` with metro preset:
```
// Ugly experimental code below
const babelLoaderRule = config.module!.rules[1];
const babelLoaderDef = (babelLoaderRule.use as webpack.RuleSetUseItem[])[0] as webpack.RuleSetLoader;
(babelLoaderDef.options! as any).presets = [[require.resolve('metro-react-native-babel-preset'), { disableImportExportTransform: false }]];

config.module!.rules.push({
  test: /\.tsx?$/,
  loader: 'ts-loader',
  options: {
    onlyCompileBundledFiles: true,
    transpileOnly: env.bundleTarget !== 'file',
    compilerOptions: {
      declaration: false,
      sourceMap: true
    }
  }
});
```

So what `babelWorkerLoader` receives as 'sources' is the compiled JS code, with a sourcemap that points back to original TypeScript file. However, `transform.js` then overwrites the result sourcemap with the compiled JS code (because it thinks it's the real source), resulting in the loss of TypeScript code from sourcemap.

By removing the overwrite, this PR preserves the original source that comes before `babelWorkerLoader`.

While our loader setup is somewhat unusual, I'm curious about the reason for the overwrite logic introduced by #689. If this PR is rejected, some explanation would be truly helpful 👍

### DIFF
--- a/packages/haul-core/src/webpack/loaders/babelWorkerLoader/vendor/transform.ts
+++ b/packages/haul-core/src/webpack/loaders/babelWorkerLoader/vendor/transform.ts
@@ -20,7 +20,7 @@ export default async function(
 
   const { ast, code, map, metadata, sourceType } = result;
 
-  if (map?.sourcesContent?.length) {
+  if (map && (!map.sourcesContent || !map.sourcesContent.length)) {
     map.sourcesContent = [source];
   }
 


statm:
cc @NoemiRozpara (original author of #689), thank you!

zamotany:
@statm `babelWorkerLoader` is heavily based on `babel-loader` and they overwrite the source maps in their transform as well: https://github.com/babel/babel-loader/blob/master/src/transform.js#L24-L26

statm:
@zamotany Interesting! I'll convert this back to a draft while doing more research.

statm:
@zamotany The TypeScript paraphrase of original babel-loader's transform.js is inaccurate.

Original `transform.js`:
```
if (map && (!map.sourcesContent || !map.sourcesContent.length)) {
  map.sourcesContent = [source];
}
```
(Fill `map.sourcesContent` if it's empty.)

Haul's `transform.ts` with [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining):
```
// TS
if (map?.sourcesContent?.length) {
  map.sourcesContent = [source];
}

// Compiled JS
if (map && map.sourcesContent && map.sourcesContent.length) {
  map.sourcesContent = [source];
}
```
(Fill `map.sourcesContent` if it's **not** empty.)

I just made another push to fix the logic.

PR #749, opened 2020-07-07
feat: refactor plugins

zamotany:
## Motivation

Refactor plugins to make them more descriptive of what they do, generic and reusable.


packages/haul-core/src/webpack/plugins/LooseModePlugin.ts
```diff
@@ -0,0 +1,76 @@
...
+      this.shouldUseLoosMode = () => false;
+    }
+  }
+
+  apply(compiler: webpack.Compiler) {
+    compiler.hooks.make.tap(
+      'LooseModeWebpackPlugin',
+      (compilation: webpack.compilation.Compilation) => {
+        compilation.moduleTemplates.javascript.hooks.render.tap(
+          'LooseModeWebpackPlugin',
```
thymikee, line 53: for consistency
```suggestion
      'LooseModePlugin',
      (compilation: webpack.compilation.Compilation) => {
        compilation.moduleTemplates.javascript.hooks.render.tap(
          'LooseModePlugin',
```

thymikee: Or maybe even prefix this with `Haul`? Not sure how webpack resolves naming conflicts, I supposed there may be various LooseModePlugins out there

packages/haul-core/src/webpack/plugins/PreloadBundlesPlugin.ts
```diff
@@ -0,0 +1,34 @@
...
+  apply(compiler: webpack.Compiler) {
+    compiler.hooks.compilation.tap('PreloadBundlesPlugin', compilation => {
+      (compilation.mainTemplate as any).hooks.bootstrap.tap(
+        'PreloadBundlesPlugin',
+        (source: string) => {
+          const preload = this.bundles.length
+            ? `${this.bundles.map(
+                bundleName =>
+                  `if (!this["${bundleName}"]) { this.bundleRegistryLoad("${bundleName}", true, true); }\n`
+              )}\n`
```
thymikee, line 27: if we have `dedent` in deps somewhere, then it could be used to make it easier to read the template, but that's just a suggestion

packages/haul-core/src/webpack/plugins/PreloadModulesDllPlugin.ts
```diff
@@ -98,13 +116,13 @@ export default class InitCoreDllPlugin {
               ...args: any[]
             ) => {
               tap.fn(compilation, ...args);
-              const initCoreDllModuleFactory = new InitCoreDllModuleFactory(
+              const initCoreDllModuleFactory = new PreloadDllModuleFactory(
                 () => {
                   const setupFilesModulesIds = compilation.modules.reduce(
```
thymikee, line 121: rename to just `modulesIds`? `setupFile` is not a thing now, so it may be confusing

packages/haul-core/src/webpack/plugins/RamBundlePlugin/RamBundlePlugin.ts
```diff
@@ -0,0 +1,308 @@
...
+              typeof webpackModule.id === 'string'
+                ? webpackModule.index
+                : webpackModule.id
+            }.js`,
+          },
+        };
+      })
+    );
+
+    minifyWorker.end();
```
thymikee, line 193: `end()` is now async: https://github.com/facebook/jest/blob/master/packages/jest-worker/src/index.ts#L140, make sure to check if the version you're on is async as well and if not, mark as todo or something

zamotany: I had a problem previosuly, when awaiting `end()` would never resolve and the whole process would hang indefinitely.

Issue #748, opened 2020-06-26
Asset naming algorithm causes ENAMETOOLONG when using react-navigation


spion:
The name chosen in:

https://github.com/callstack/haul/blob/c0bfea756dfd52f965531e0b92927fcda8440980/packages/haul-core/src/webpack/loaders/assetLoader.ts#L50-L56

causes ENAMETOOLONG due to basename length over 255 characters:

```
error ▶︎ [Error: ENAMETOOLONG: name too long, open '/.../drawable-mdpi/node_modules_pnpm_reactnavigation_stack551_a182da6647d2184b7d281770a34a973a_node_modules_reactnavigation_stack_src_views_assets_backiconmask.png'] { errno: -36, code: 'ENAMETOOLONG', syscall: 'open', path: '/.../drawable-mdpi/node_modules_pnpm_reactnavigation_stack551_a182da6647d2184b7d281770a34a973a_node_modules_reactnavigation_stack_src_views_assets_backiconmask.png' }
```

PR #752, opened 2020-07-11
Fix issue with long asset names generated from pnpm paths

spion:
### Summary

This PR fixes issue #748 

### Test plan

Example output when compiling react-navigation in a pnpm project:

```
68502 % ls drawable-mdpi/  
gation_stack551_a182da6647d2184b7d281770a34a973a_node_modules_reactnavigation_stack_src_views_assets_backiconmask_3.png
m_reactnative0622_686ec919160581136b28b7fe63160a42_node_modules_reactnative_libraries_logbox_ui_logboximages_loader_6.png
native0622_686ec919160581136b28b7fe63160a42_node_modules_reactnative_libraries_logbox_ui_logboximages_alerttriangle_7.png
native0622_686ec919160581136b28b7fe63160a42_node_modules_reactnative_libraries_logbox_ui_logboximages_chevronright_4.png
navigation_stack551_a182da6647d2184b7d281770a34a973a_node_modules_reactnavigation_stack_src_views_assets_backicon_2.png
pm_reactnative0622_686ec919160581136b28b7fe63160a42_node_modules_reactnative_libraries_logbox_ui_logboximages_close_1.png
```

Might still be an issue if multiple processes or worker threads are running an asset loader module in parallel. Let me know if you'd like a different approach

### DIFF
--- a/packages/haul-core/src/webpack/loaders/assetLoader.ts
+++ b/packages/haul-core/src/webpack/loaders/assetLoader.ts
@@ -15,6 +15,9 @@ type Config = {
   publicPath?: string | ((path: string) => string);
 };
 
+const MAX_FILENAME_LENGTH = 120;
+let longNameAssetCount = 1;
+
 async function assetLoader(this: any) {
   this.cacheable();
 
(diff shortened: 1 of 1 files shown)


zamotany:
Thanks for the PR @spion
Maybe instead of slicing the string and adding a counter to it, we could hash the full long path into the eg 32 character hash? To simplify the logic we could do this in all cases, not only if the name is longer than 255. Do you know how Metro does this? Or if they are using a hash for assets??

spion:
I think metro is completely incapable at working with pnpm due to its extensive usage of symlinks.

What about hashing the full path but still using the basename? I'd imagine a lot of people will appreciate having at least some indication of what the file is about :grinning: 

P.S. Any suggestions about which hash to use?

Issue #756, opened 2020-10-21
Support other browsers (eg: user-chosen defaults)


evancharlton:
## Describe the feature

Currently, when using the `@haul-bundler/cli start [..]`, Chrome is opened to launch the debugger. This is wonderful! However, my default browser is not Chrome, so it's jarring to have Chrome opened for this suddenly.

On this machine, my default browser is Edge. Edge is [now Chromium-based](https://www.theverge.com/2020/1/15/21066767/microsoft-edge-chromium-new-browser-windows-mac-download-os) and [supports native Chrome extensions](https://www.howtogeek.com/411830/how-to-install-google-chrome-extensions-in-microsoft-edge/). It would be great to have a way to respect platform defaults rather than making assumptions.

## Motivation

I try to maintain a separation between personal browsers & work browsers, and having the dev tools automatically make a choice complicates things for me 😁 

## Related Issues

I was unable to find previous requests for this.

PR #757, opened 2020-10-21
feat: Allow browser to be specified

evancharlton:
Instead of making the assumption on behalf of the user, allow them to
control the browser which is used for the debugger UI. This is
controlled via the `RCT_DEBUGGER_BROWSER` environment variable. If this
variable is populated, then it will be tried first, before falling back
to the baked-in defaults.

### Test plan

This can be tested by setting `RCT_DEBUGGER_BROWSER` to another browser
and then running the `start` command for the `@haul-bundler/cli`
package.

## Issues

This addresses issue #756

### DIFF
--- a/packages/haul-core/src/server/launchBrowser.ts
+++ b/packages/haul-core/src/server/launchBrowser.ts
@@ -12,11 +12,11 @@ export default function launchBrowser(runtime: Runtime, url: string) {
    * Run Chrome (Chrome Canary) or supported platform.
    * In case of macOS, we can eventually fallback to Safari.
    *
-   * select(attemp1, attemp2, attemp3,...) // attempt to run is from left to right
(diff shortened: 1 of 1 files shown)


evancharlton:
I haven't looked into how to test this in an automated way, but I can if it agreed that this is a valid approach.

packages/haul-core/src/server/launchBrowser.ts
```diff
@@ -28,8 +28,17 @@ export default function launchBrowser(runtime: Runtime, url: string) {
     {
       // No Canary / Chrome, let's run Safari
       darwin: openWith('safari'),
-    }
-  ).catch((error: Error) => {
+    },
+  ];
+
+  // If the user has provided a specific browser preference, try that one first.
```
evancharlton, line 34: There's certainly an argument to be made for **not** doing a fallback if the user has specified a preference.

Issue #746, opened 2020-06-26
Module.createRequireFromPath() is deprecated.


sraka1:
## Environment

Node v13.8.0

## Description

Module.createRequireFromPath() is deprecated since Node v12.2.0. An error is thrown when it is used on v13.

## Reproducible Demo

Use haul in a project with Node version > 13.

## Proposed fix

In https://github.com/callstack/haul/blob/d64efa45b30cd92026c9b5a595ed838db038b077/packages/haul-core/src/utils/importModule.ts#L88 something like the following could be implemented (to main compatibility with Node 10-12)

```
  const createRequireWrapper = Module.createRequire || Module.createRequireFromPath
  // Create resolver for this module.
  const currentResolve = ((createRequireWrapper(
    module.filename
  ) as unknown) as {
    resolve: RequireResolve;
  }).resolve;
```

The method seems to be identical in function, only its naming has changed.

PR #758, opened 2020-11-03
Support node 14

eivisand:
### Summary
Solves [Module.createRequireFromPath() is deprecated](https://github.com/callstack/haul/issues/746)
Node 14 has made it to long term support, and it crashes when used with Haul. To fix this I've bumped `@types/node` to currtent, and added `Module.createRequire` with fallback to `Module.createRequireFromPath` where only `Module.createRequireFromPath` was used before.
In addition I fixed type errors that came with the `@types/node` update.

### Test plan
I'm not sure how best to test, I ran the jest tests locally, and failed 7 test suites, 10 tests, and 3 snapshots on the PR branch vs 10 test suites, 15 tests and 3 snapshots on master. Looks like createRequireFromPath and createRequire are equivalent api equivalent, apart from createRequire supporting URL input, so they should be swappable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "@types/lodash.merge": "^4.6.6",
     "@types/lodash.throttle": "^4.1.6",
     "@types/mkdirp": "^0.5.2",
-    "@types/node": "^10.17.13",
+    "@types/node": "^14.14.6",
     "@types/node-fetch": "^2.5.4",
     "@types/npm-registry-fetch": "^4.0.1",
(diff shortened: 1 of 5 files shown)


eivisand:
Pinging this, would be nice with a review 🌝

tido64:
@zamotany, @thymikee: Is this something you could take a look at?

thymikee (approved):
@zamotany please take a look and merge if ready

